PR #5, opened 2019-07-29
simplify: only use NodeRef, no more need for ElementRef

mpollmeier:
inlcudes https://github.com/ShiftLeftSecurity/overflowdb/pull/4 - merge that first


tinkerpop3/src/main/java/io/shiftleft/overflowdb/structure/NodeRef.java
```diff
@@ -5,28 +5,112 @@
...
+ * When it's cleared, it will be persisted to an on-disk storage.
+ */
+public abstract class NodeRef<N extends OverflowDbNode> implements Vertex {
+  public final long id;
 
-  public NodeRef(final Object vertexId, final Graph graph, V vertex) {
-    super(vertexId, graph, vertex);
+  protected final OverflowDbGraph graph;
+  protected N node;
+  private boolean removed = false;
```
ml86: Do we still need the removed field? I think that is a left over from tinker world right?

mpollmeier: true, good point, done

PR #6, opened 2019-07-30
no need for NodeRefWithLabel - remove

mpollmeier:
never hold label as a member, not even in refs


ml86:
This seems to also contain a lot of renaming. Was it not possible to split that?

PR #36, opened 2019-09-09
Added PackedIntArray for OdbNode.edgeOffsets

lifecoder:


lifecoder:
@mpollmeier I've added docs and a unit test

PR #41, opened 2019-10-02
clean up ordering of indices, offer fast neighbor access api

bbrehm:
This PR introduces a new API to grab neighbors by integer. The `.vertices(labelstring, directionstring)` calls are on the fast path, hence hoisting the multi-iterator and hashmap stuff can save some time.

### DIFF
--- a/tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
+++ b/tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
@@ -31,6 +31,9 @@ public class NodeLayoutInformation {
   /* possible edge property keys, grouped by edge label.
    * n.b. property keys are of type `HashSet` (rather than just `Set`) to ensure `.size` has constant time */
   private final Map<String, HashSet<String>> edgePropertyKeysByLabel;
+  /* maps offsetPos -> number of edge properties
+  */
+  public final int[] nPropertiesbyOffsetPos;
 
   /* position in stride (entry within `adjacentNodesWithProperties`) for a given edge label and edge property key
    * 1-based, because index `0` is the adjacent node ref */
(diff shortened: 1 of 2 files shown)


tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
```diff
@@ -31,6 +31,9 @@
   /* possible edge property keys, grouped by edge label.
    * n.b. property keys are of type `HashSet` (rather than just `Set`) to ensure `.size` has constant time */
   private final Map<String, HashSet<String>> edgePropertyKeysByLabel;
+  /* maps offsetPos -> number of edge properties
+  */
+  public final int[] nPropertiesbyOffsetPos;
```
mpollmeier: this should be private, so that users don't change the contents

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
```diff
@@ -51,19 +54,25 @@ public NodeLayoutInformation(Set<String> propertyKeys,
...
      * sort them by edge label to ensure we get the same offsets between restarts
      * n.b. this doesn't support schema changes */
+
+    allowedOutEdgeLabels = new String[outEdgeLayouts.size()];
+    allowedInEdgeLabels = new String[inEdgeLayouts.size()];
     int offsetPosition = 0;
+    nPropertiesbyOffsetPos = new int[outEdgeLayouts.size() + inEdgeLayouts.size()];
+    int i = 0;
     outEdgeToOffsetPosition = new HashMap<>(outEdgeLayouts.size());
-    for (EdgeLayoutInformation edgeLayout : sortByLabel(outEdgeLayouts)) {
```
mpollmeier: Sorting the layouts ensures that we get consistent layouts in between jvm starts. 
If you stop sorting it by label we get a different layout every time we start, i.e. we cannot save to disk and load it back in a new jvm instance.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
```diff
@@ -388,12 +388,16 @@ protected void removeEdge(Direction direction, String label, int blockOffset) {
...
+
+  // Simplify hoisting of string lookups.
+  // This is on the hot path, hence final.
+public final Iterator<NodeRef> createAdjacentNodeIteratorByOffSet(int offsetPos){
     if (offsetPos != -1) {
       int start = startIndex(offsetPos);
       int length = blockLength(offsetPos);
-      int strideSize = getStrideSize(label);
-
+      int strideSize = layoutInformation().nPropertiesbyOffsetPos[offsetPos] + 1;
```
mpollmeier: why only use that optimization here, rather than directly in `getStrideSize`, so that other usages benefit from it as well?

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
bbrehm: The layouts are passed in a List, which is not a set/hashmap. The new rule is: We use the same ordering as the caller in the constructor. Is that acceptable?

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
bbrehm: The idea of this PR is to move all strings (e.g. label) off the hot path. For this I just needed a fast offsetPos->stride map, because the fast API use knows nothing about labels/strings and only knows offsetPos.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
bbrehm: OdbNode class needs to access it.

mpollmeier:
```
private final int[] nPropertiesbyOffsetPos;

public final int getPropertyCountByOffsetPos(int offsetPos) { return [nPropertiesbyOffsetPos]; }
```
that way users can't shoot themselves in the foot too easily

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
mpollmeier: true, looking at the implementation now, seems ok to not sort again.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
mpollmeier: ok

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
mpollmeier: edit: i updated my proposal above with an (imo) better name, which also let's you drop the comment

tinkerpop3/src/main/java/io/shiftleft/overflowdb/NodeLayoutInformation.java
```diff
@@ -51,19 +54,25 @@ public NodeLayoutInformation(Set<String> propertyKeys,
     /* create unique offsets for each edge type and direction
```
ml86: Please also update the comments. This does not seem to be true anymore.

PR #44, opened 2019-10-07
Add serialization support for chars

### LINKED ISSUE #42: java.lang.Character cannot be serialized
I would need to create a node that has a property of type "java.lang.Character", but when OverflowDB tries to persist it to disk, I get the error message that this property type is not implemented

Masrepus:
Should fix #42 
Unfortunately I couldn't get it to create a correct jar for me (no idea how jar publishing with sbt works), so couldn't test it in action yet...


Masrepus:
@mpollmeier is this something that could be merged or is this not desired?

mpollmeier:
Oh sorry I totally missed this PR. sure that looks good, thank you!
This will get released automatically by https://travis-ci.org/ShiftLeftSecurity/overflowdb
For next time, to publish a jar locally with sbt: `sbt publishLocal`

Masrepus:
Ah thanks for the tip!

PR #46, opened 2019-10-07
Versionfix

bbrehm:
Yet another attempt to make travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: scala
-dist: bionic
+dist: xenial
 jdk: openjdk8
 if: tag IS blank
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: scala
 dist: bionic
-jdk: openjdk11
+jdk: openjdk8
 if: tag IS blank
 
 before_install:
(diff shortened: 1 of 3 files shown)


bbrehm:
Cross compiling openjdk11 -> java 1.8 appears to be harder than expected. So let's just use openjdk8.

Unfortunately, the tests don't cover the publish part, so there is AFAIK no way of knowing what travis will emit until we merge this. Locally everything runs fine even before this commit.

@mpollmeier Can we make travis publish every single time it attempts a build (with version = commit hash)? That would simplify everything.

mpollmeier:
sbt-ci-release-early does publish a release every single time it builds, the only limitation is that it's only doing that on master. is that what you want to change? i'd like to keep the process and plugin straightforward

bbrehm:
Yes, I'd like a no-name (i.e. commit hash) release every time a build is triggered for any reason.

This would be very nice when handling changes that affect multiple repos simultaneously / feature branches. This means that all tests can pass non-locally in all repos before the first PR is merged.

mpollmeier:
if you would like to experiment with the idea, here's a good place to start: this is where the 'release' commands are implemented. 

https://github.com/ShiftLeftSecurity/sbt-ci-release-early/blob/master/src/main/scala/ci/release/early/Plugin.scala#L24-L44

You could add your own command (`ci-release-nonmaster`). It's an sbt build as well, you you can run `sbt publishLocal` and use that locally built version in another build.

PR #50, opened 2019-10-15
Add ability to disable overflow manager

lifecoder:


tinkerpop3/src/test/java/io/shiftleft/overflowdb/storage/GraphSaveRestoreTest.java
```diff
@@ -30,7 +30,7 @@ public void greenField() throws IOException {
     final Long vertex0Id;
     final Long vertex1Id;
     // create graph and store in specified location
-    try (OdbGraph graph = newGratefulDeadGraph(overflowDb, false)) {
+    try (OdbGraph graph = newGratefulDeadGraph(overflowDb, true)) {
```
mpollmeier: why would you enable the overflow for these tests?

lifecoder: they did not pass because when reference manager does not collect node refs it did not save anything on close

mpollmeier: ah i see, ok then as discussed try to separate the two so that this will work without overflow enabled

PR #62, opened 2020-01-20
refactor index implementation

mpollmeier:
* not generic - it's only for NodeRefs
* avoid usage of static methods
* simplify implementation, remove redundant fields etc


tinkerpop3/src/test/java/io/shiftleft/overflowdb/IndexesTest.java
```diff
@@ -25,7 +23,7 @@ public void shouldUseIndices() throws IOException {
 
     { // tests with index
       OdbGraph graph = GratefulDead.newGraphWithData();
-      graph.createIndex("performances", Vertex.class);
+      graph.indexManager.createNodePropertyIndex("performances");
```
lifecoder: Not sure if indexes are used somewhere in dependent projects, maybe this will require additional renaming? Otherwise looks OK

mpollmeier: yes i think they are, will provide follow up PRs on those

PR #64, opened 2020-01-22
Stored indexes

lifecoder:


tinkerpop3/src/test/java/io/shiftleft/overflowdb/IndexesTest.java
```diff
@@ -75,5 +77,36 @@ public void shouldUseIndicesCreatedBeforeLoadingData() throws IOException {
...
+      assertEquals(142, (long) g.V().has("performances", P.eq(1)).count().next());
+      avgTimeWithStoredIndex = TimeUtil.clock(loops, () -> g.V().has("performances", P.eq(1)).count().next());
+      graph.close();
+    }
+
+    System.out.println("avgTimeWithIndexCreation = " + avgTimeWithIndexCreation);
+    System.out.println("avgTimeWithStoredIndex = " + avgTimeWithStoredIndex);
+    assertTrue("avg time with index should be (significantly) less than without index",
+        avgTimeWithIndexCreation > avgTimeWithStoredIndex);
+  }
```
mpollmeier: Please also add a test that asserts the presence of an index in an way that not dependent on the environment (but keep this one, to ensure we can also test it end to end). 

I mean something like 
* setup odb with index
* close, restart
* graph.indexManager.getIndex("abc").count shouldBe 50

That one we don't need to leave as 'ignored' and it'll be easier to debug, in case it goes red.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbGraph.java
```diff
@@ -137,10 +138,25 @@ private void initElementCollections(OdbStorage storage) {
     }
 
     currentId.set(maxId + 1);
+    initializeStoredIndices(storage);
     long elapsedMillis = System.currentTimeMillis() - start;
     logger.info("initialized " + this.toString() + " from existing storage in " + elapsedMillis + "ms");
   }
 
+  private void initializeStoredIndices(OdbStorage storage) {
```
mpollmeier: I introduced the `IndexManager` in the last PR and was hoping that we could move all index related functionality there - this applies to all changes in this file.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbGraph.java
```diff
@@ -137,10 +138,25 @@ private void initElementCollections(OdbStorage storage) {
...
+    storage
+        .getIndexNames()
+        .stream()
+        .forEach(indexName -> loadIndex(indexName, storage));
+  }
+
+  private void loadIndex(String indexName, OdbStorage storage) {
+    assert indexName != null;
```
mpollmeier: You're probably aware, but asserts in java aren't checked at runtime unless the `-ea` flag is set.
I'd just drop that line - if anyone passes `null` they deserve to suffer.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
```diff
@@ -98,13 +101,70 @@ public NodeSerializer getNodeSerializer() {
...
+        .map(s -> {System.out.println("filtered name: " + s); return s;})
+        .collect(Collectors.toConcurrentMap(s -> removeIndexPrefix(s), s -> s));
+  }
+
+  public Set<String> getIndexNames() {
+    return getIndexNameMap(mvstore).keySet();
+  }
+
+  private String removeIndexPrefix(String s) {
+    assert s.startsWith(INDEX_PREFIX);
```
mpollmeier: same remark about `assert` as above applies

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
```diff
@@ -12,8 +13,10 @@
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class OdbStorage implements AutoCloseable {
```
mpollmeier: IMO the readability would benefit if we move all purely index-related code into a separate unit/section that's wired in, e.g. a as a abstract class, or a separate class member. There's many ways, but the goal is: keep this class small and readable.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
```diff
@@ -98,13 +101,70 @@ public NodeSerializer getNodeSerializer() {
...
+    getIndexNames().forEach(this::clearIndex);
+  }
+
+  private void clearIndex(String indexName) {
+    System.out.println("clearing index: " + indexName);
+    openIndex(indexName).clear();
+  }
+
+  public void saveIndex(String propertyName, Map<Object, Set<NodeRef>> indexMap) {
+    final MVMap<Object, long[]> indexStore = openIndex(propertyName);
```
mpollmeier: worth checking: are there any limitations on mvmap names, so that we'd need to adjust/encode the names?

lifecoder: I've checked the code and it does not look that there are any limitations

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
lifecoder: Yes, I'll remove all asserts, they are just a leftover after I was debugging it :)

tinkerpop3/src/test/java/io/shiftleft/overflowdb/IndexesTest.java
lifecoder: Done

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
lifecoder: Moved almost everything to OdbIndexManager

tinkerpop3/src/test/java/io/shiftleft/overflowdb/IndexesTest.java
```diff
@@ -75,5 +79,57 @@ public void shouldUseIndicesCreatedBeforeLoadingData() throws IOException {
...
+        assertEquals(142, (long) g.V().has("performances", P.eq(1)).count().next());
+      }
+    });
+
+    System.out.println("avgTimeWithIndexCreation = " + avgTimeWithIndexCreation);
+    System.out.println("avgTimeWithStoredIndex = " + avgTimeWithStoredIndex);
+  }
+
+  @Test
+  public void shouldStoreIndexes() throws IOException {
```
mpollmeier:
```suggestion
  public void shouldStoreAndRestoreIndexes() throws IOException {
```

PR #65, opened 2020-02-10
revive dirty flag to ensure we're only serializing when necessary

mpollmeier:
* also handle for multiple threads
* added synchronization to a bunch of other locations
* added `final` to a bunch of methods to help JITc
* multi threading scenario1: node is updated while it's read from storage in another thread: handled by synchronization in NodeRef.get
* multi threading scenario2: node or edge is updated while it's serialized
  * on close: not a problem because graph is immutable
  * on overflow: ok because serializer marks it as clean *before* it starts and updating the node marks it as dirty *after* it starts

### DIFF
--- a/tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
+++ b/tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
@@ -28,17 +28,10 @@ import java.util.stream.StreamSupport;
  * Node that stores adjacent Nodes directly, rather than via edges.
  * Motivation: in many graph use cases, edges don't hold any properties and thus accounts for more memory and
  * traversal time than necessary
- *
- * TODO: remove tinkerpop dependency
  */
 public abstract class OdbNode implements Vertex {
 
   public final NodeRef ref;
-
-  @Override
-  public String label() {
-    return ref.label();
-  }
 
   /**
    * holds refs to all adjacent nodes (a.k.a. dummy edges) and the edge properties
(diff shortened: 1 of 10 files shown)


tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
```diff
@@ -448,10 +463,13 @@ private int storeAdjacentNode(Direction direction,
       }
     }
 
+    /* marking as dirty *after* we updated - if node gets serialized before we finish, it'll be marked as dirty */
+    this.markAsDirty();
```
lifecoder: @mpollmeier There is also `setEdgeProperty` which is called when you set edge property from this fn and also from the OdbEdge through `property` method. I think it should also mark the node as dirty.

lifecoder:
> That was quick. Can you speak a bit about the increase in memory usage of this a bit for the record? Each flag has that field `dirty` now, correct? How do we store it, how large is it?
@fabsx00 This should be 4 bytes per node and 8 bytes on very large heap size.
We talked about making something like bitset or a bloom filter in the future, but this is just a quick solution for now.

mpollmeier:
> This should be 4 bytes per node and 8 bytes on very large heap size.

And notice that we're talking about nodes here, not edges (since they only exist virtually).

tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
mpollmeier: we do that already: https://github.com/ShiftLeftSecurity/overflowdb/blob/3fe462229187fd5a0ae908d1f501c437745c62a1/tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java#L228

mpollmeier:
@fabsx00 we could replace the dirty flag with a bitset on the graph or a bloom filter (that allows entry removal) to save some space, but we thought it's probably not be worth it for now. Thoughts?

PR #67, opened 2020-02-11
Parallel node serialization.

lifecoder:


lifecoder:
> This PR is called 'parallel node serialization', but isn't this doing just the opposite, i.e. we previously serialized them in parallel, but now it's doing it sequentially in a single thread executor?
> 
> I'll approve to not block you, but please clarify and add a more detailed description of why this is better to the PR description before merging.
It does serialization in parallel but writes to the mvstore using single thread.
More appropriate name would be that it makes access to mvstore single threaded

PR #69, opened 2020-02-11
trim overallocations, intern strings in deserialization, improve timings

bbrehm:
This adds API for trimming overallocations. Intention is to run this as a final enhancement pass. @mpollmeier do we need to set a dirty flag on this?

Further, the timing code was silly (time in milliseconds for node deserialization -- if this is average > 1 ms, then we can throw away the code and start again). However, do we really want these semantics? I.e. do we really want to exclude the time for extracting the `bytes[]` buffer from the timing? Speed of nanoTime needs to be tested in our actual deployment, because it may well depend on whatever $cloud_provider is doing. As commented, I would prefer a small assembly file that we build into shared library and call with JNI (afaiu we only care about x86_64 with linux / windows / macos).

Further, added string interning on deserialization. Somebody needs to intern these strings, and g1gc only interns the buffers, not the String objects.

This ain't threadsafe, but deserialization should not be multithreaded anyway.

### DIFF
--- a/tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
+++ b/tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbNode.java
@@ -571,4 +571,28 @@ public abstract class OdbNode implements Vertex {
       throw new IllegalArgumentException("specializedEdgeFactory for label=" + label + " not found - please register on startup!");
     return edgeFactory.createEdge(ref.graph, outNode, inNode);
   }
+  /**
+   * Trims the node to save storage: shrinks overallocations
+   * */
+  public int trim(){
+    int newSize = 0;
+    for(int offsetPos = 0; 2*offsetPos < edgeOffsets.length(); offsetPos++){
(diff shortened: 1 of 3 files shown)


bbrehm:
>Deserializing multiple nodes in parallel should be fine though.

Damn, then we'll have to change the string internalization: `THashMap` is not threadsafe. Afaik ConcurrentHashMap is one of the better alternatives out there, but this will probably be much slower :(

>deserializationTimeSpent is just for debugging - changing it to nanos is good, but additionally let's hide those measurements behind a config option (off by default). You can add it to this PR or I do it in a separately - as you please.

I'll leave that to you and read your PR, since I'm not up to date with how we configure odb.

bbrehm:
@mpollmeier @lifecoder Can you take a look at the non-reproducible travis fail? I.e. think whether that is an obvious race condition?

mpollmeier:
> Can you take a look at the non-reproducible travis fail?

Looks like my bad, will TAL. 
Also, will move the measurement of deserialization time behind a feature flag.

PR #70, opened 2020-02-13
hide serialization stats recording behind flag (off by default)

mpollmeier:


tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/NodeDeserializer.java
```diff
@@ -22,31 +22,25 @@
...
     String interned = interner.putIfAbsent(s, s);
     return interned == null ? s : interned;
   }
 
-
   public final OdbNode deserialize(byte[] bytes) throws IOException {
-    // todo: only time when some config is set
-
-    long start = System.nanoTime();
+    long startTimeNanos = System.nanoTime();
```
bbrehm: We would probably need `long startTimeNanos = statsEnabled ? System.nanoTime() : 0;`.

The point is that `System.nanoTime()` is afaiu a syscall on linux+openjdk8 (relatively fast, but this depends a lot on system config). This means that it might be surprisingly expensive, depending on setup. We absolutely want to prevent a setting where the message deserialization takes 300 ns, and we pay an extra 400 ns for a roundtrip to the kernel, just to collect stats we don't even collect. 

If this was an `RDTSC`, I wouldn't care (24 cycles on broadwell); but `System.nanoTime()` makes a lot of expensive extra guarantees.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/NodeSerializer.java
```diff
@@ -5,22 +5,19 @@
...
-  private int serializedCount = 0;
-  private long serializationTimeSpentMillis = 0;
+public class NodeSerializer extends BookKeeper {
+  public NodeSerializer(boolean statsEnabled) {
+    super(statsEnabled);
+  }
 
   public byte[] serialize(OdbNode node) throws IOException {
-    long start = System.currentTimeMillis();
+    long startTimeNanos = System.nanoTime();
```
bbrehm: Same as above: Only ask the kernel for the time if we actually care.

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/NodeDeserializer.java
mpollmeier: good point, yup makes sense

PR #71, opened 2020-02-13
fix flakey tests by using synchronized bookkeeping

mpollmeier:


mpollmeier:
includes https://github.com/ShiftLeftSecurity/overflowdb/pull/70

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/BookKeeper.java
```diff
@@ -0,0 +1,32 @@
...
+  protected final Logger logger = LoggerFactory.getLogger(getClass());
+  public final boolean statsEnabled;
+  private AtomicInteger totalCount = new AtomicInteger(0);
+  private AtomicLong totalTimeSpentNanos = new AtomicLong(0);
+
+  protected BookKeeper(boolean statsEnabled) {
+    this.statsEnabled = statsEnabled;
+  }
+
+  protected void recordStatistics(long startTimeNanos) {
```
bbrehm: We can do it that way, but the statistics will be very flawed on multithreaded: All threads try to write to the same counter, and the poor cpu chokes trying to synchronize that cache line between all the cores. If we are OK with that, then sure, it's a debug feature that doesn't need to be done right. Alternative would be thread-local storage.

PR #73, opened 2020-02-13
Add dirty flag for indexes

lifecoder:


tinkerpop3/src/main/java/io/shiftleft/overflowdb/OdbIndexManager.java
```diff
@@ -45,12 +48,14 @@ private void checkPropertyName(String propertyName) {
   }
 
   public final void loadNodePropertyIndex(final String propertyName, Map<Object, long[]> valueToNodeIds) {
+    dirtyFlags.put(propertyName, false);
```
mpollmeier: shouldn't this happen _after_ the index is loaded, i.e. at the end of this method?

PR #74, opened 2020-03-11
ensure temp storage file is deleted on `close`

mpollmeier:
i.e. unless persistent storage is explicitly requested, in the form of
the user specifying a storage location

### DIFF
--- a/tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
+++ b/tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
@@ -21,6 +21,7 @@ public class OdbStorage implements AutoCloseable {
   protected final Optional<NodeDeserializer> nodeDeserializer;
 
   private final File mvstoreFile;
+  private final boolean doPersist;
   private MVStore mvstore; // initialized in `getNodesMVMap`
   private MVMap<Long, byte[]> nodesMVMap;
   private boolean closed;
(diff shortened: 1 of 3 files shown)


mpollmeier:
cc @lifecoder

tinkerpop3/src/test/java/io/shiftleft/overflowdb/OverflowTest.java
```diff
@@ -3,6 +3,7 @@
 import io.shiftleft.overflowdb.testdomains.simple.TestNode;
 import io.shiftleft.overflowdb.testdomains.simple.SimpleDomain;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.junit.Test;
```
hubertp: ?

tinkerpop3/src/main/java/io/shiftleft/overflowdb/storage/OdbStorage.java
```diff
@@ -95,10 +98,11 @@ public void close() {
     closed = true;
     logger.info("closing " + getClass().getSimpleName());
     if (mvstore != null) mvstore.close();
+    if (!doPersist) mvstoreFile.delete();
```
hubertp: wondering if this should be maybe in try/catch?

mpollmeier: if the tmp file can't get deleted, closing the storage is incomplete and we should escalate that up. 
what would you suggest we do in the catch block?

PR #118, opened 2020-07-02
repeat step similar (but superior to) tinkerpop's

mpollmeier:
It's superior to Tinkerpop's repeat because:
* clearer syntax: tinkerpop uses arbitrary emit/repeat/until combinations which have semantic differences due to their order
* works for domain-specific traversals as well as generic graph traversals
* supports DFS and BFS (tp says they use DFS, but actually does use BFS)
* supports arbitrary number of iterations (tp stalls)

from the scaladoc: 
Repeat the given traversal

The @param behaviourBuilder allows you to configure 1) when the repeat step will end (times/until),
2) whether it should emit elements it passes by, and 3) which search algorithm to use (depth-first or breadth-first).

By default it will continue repeating until there's no more results, not emit anything along the way, and use
depth first search, i.e. go deep before wide.

Here are some popular alternative behaviour configurations:
```
.repeat(_.out)(_.times(3))                               // perform exactly three repeat iterations
.repeat(_.out)(_.until(_.property(Name).endsWith("2")))  // repeat until the 'Name' property ends with '2'
.repeat(_.out)(_.emit)                                   // emit everything along the way
.repeat(_.out)(_.emit.breadthFirstSearch)                // emit everything, use BFS
.repeat(_.out)(_.emit(_.property(Name).startsWith("L"))) // emit if the 'Name' property starts with 'L'
```
See RepeatTraversalTests for more examples.

Note that this works for domain-specific steps as well as generic graph steps - for details please take a look at
the examples in RepeatTraversalTests: both {{{.followedBy}}} and {{{.out}}} work.


traversal/src/main/scala/overflowdb/traversal/RepeatStep.scala
```diff
@@ -0,0 +1,117 @@
...
+   * See RepeatTraversalTests for more detail (and a test for the above).
+   *
+   * Note re implementation: using recursion results in nicer code, but uses the JVM stack, which only has enough space for
+   * ~10k steps. So instead, this uses a programmatic Stack which is semantically identical.
+   * The RepeatTraversalTests cover this case.
+   */
+  object DepthFirst {
+    case class StackItem[A](traversal: Traversal[A], depth: Int)
+
+    def apply[B](repeatTraversal: B => Traversal[B], behaviour: RepeatBehaviour[B]): B => Traversal[B] = {
```
ml86, line 27: I am kind of surprised to see two completely different implementations for DepthFirst and BreadthFirst traversal. I would have expected the difference to only be the collection type for the "worklist" which would be a stack for depth first and a FIFO for breadth first.

mpollmeier: Yes good point, this is mostly an artifact of my journey to the final version. Initially the two versions looked very similar, but the DFS version used recursion, i.e. the worklist stack was the JVM call stack. Until I realised that that's not a good idea because it quickly blows up. 
Will TAL tomorrow to see if it's worthwhile to bring the two impls closer together. Thanks!

mpollmeier: yup you're right. i'll still merge this as is to have this version archived in the git history, will send a follow up pr then to unify the two options

PR #124, opened 2020-07-07
rename nodesByLabel -> nodes

mpollmeier:
node ids are of type `Long`, so `nodes(String)` is pretty obvious to be
filtering on the label


mpollmeier:
I'd regard the two on different levels: the domain specific language sits on top and can make their own choices as it best fits for that specific domain. On this level I think plural is appropriate.

PR #133, opened 2020-07-20
replace trove collection with custom list for performance and memory usage

mpollmeier:
* motivation: graph.nodes is extremely slow. while it's recommended to not traverse _all_ nodes normally, sometimes it's necessary
* IteratorUtils has an unnecessary performance hit
* iterating all nodes is 5-10x faster (only performed very simplistic benchmarks)
* less memory usage when only adding nodes
* automatically compacts so that far less memory is used when nodes are being removed

### DIFF
--- a/tinkerpop3/src/main/java/overflowdb/NodeRef.java
+++ b/tinkerpop3/src/main/java/overflowdb/NodeRef.java
@@ -299,4 +299,8 @@ public abstract class NodeRef<N extends OdbNode> implements Vertex, Node {
 
   // delegate methods end
 
+  @Override
+  public String toString() {
+    return getClass().getName() + "[label=" + label() + "; id=" + id + "]";
+  }
 }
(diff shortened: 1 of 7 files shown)


bbrehm:
Can you post the benchmarks?

The THashmap iterator code doesn't look too catastrophic to me. That is, we have one expected branch mispredict per iteration, which is suboptimal -- but when we see what the exact issue is, we probably can do better than adding so much extra code.

mpollmeier:
@bbrehm I'll merge this now, but will share the benchmarks shortly. if you find a way for us to keep using trove's collections, that's ok, but they also don't have a compaction built-in. 
n.b. 1) the "benchmarks" don't deserve their name, but are sufficient to show that there's a big performance difference
n.b. 2) part of the performance improvements comes from not using IteratorUtils any more

PR #138, opened 2020-07-24
Modify how contained_by works

bbrehm:
Central discussion thread is https://github.com/ShiftLeftSecurity/codescience/issues/4038

### DIFF
--- a/tinkerpop3/src/main/java/overflowdb/storage/NodeDeserializer.java
+++ b/tinkerpop3/src/main/java/overflowdb/storage/NodeDeserializer.java
@@ -141,6 +141,7 @@ public class NodeDeserializer extends BookKeeper {
   protected final Object[] toTinkerpopKeyValues(Map<String, Object> properties) {
     List keyValues = new ArrayList(properties.size() * 2); // may grow bigger if there's list entries
     for (Map.Entry<String, Object> entry : properties.entrySet()) {
+      //todo: We fail to properly intern strings contained in a List.
       final String key = intern(entry.getKey());
       final Object property = entry.getValue();
       keyValues.add(key);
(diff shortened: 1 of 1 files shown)


mpollmeier:
this is just WIP right?

bbrehm:
I will delete the now commented-out code before merging. However, you should tell me if I missed some other changes that are required due to this change.

mpollmeier:
ok I didn't pay much attention up until now because this looked too much like wip

Anyway, yes this should be ok as is, even if it's still using the tinkerpop api. If I port that over to the odb api now or later shouldn't really matter (as long as all tests are green obvsly).

ml86:
Why dont we need special handling for list properties anymore?

bbrehm:
There is some nebulous gremlin tinkerpop API for modifying graphs. This nebulous API appears to require this special handling (?)

We know and control every single call to all the modifications: The are done by the generated code, and triggered by diffgraph application, proto deserialization and overflowdb loading. As long as we keep our stuff compatible and don't trigger the JVM's type-errors, we can pass whatever we want where ever we want and have it mean what ever we want. So we don't care about interoperability with anything else for graph modification (we're still re-using the upstream vtable entries, though).

So I changed how the API works for us: Properties with list cardinality used to be passed one after the other, and afaiu used to have limited guarantees on the order of results; semantically, they were appended.

Now they are passed as a List, and always overwrite the previous entry and always preserve order.

mpollmeier:
Yup that analysis is correct. we used to go via the tinkerpop api to handle the lists, adding one item after the other. But since the NodeDeserializer always wanted the entire list, there wasn't really a good reason for it anyway. 
Now we're just setting the entire list at once, which is good. 

*However*: note that we now don't intern Strings that are part of a List any more... not sure from the top of my head what the best way to handle those are. Maybe @bbrehm can take a look first and I can revisit this later when I have more time.

PR #156, opened 2020-09-08
drop tricky varargs

mpollmeier:
P.within/without with varargs is dangerous, because of the unspecific
type parameter A.

example for misuse:
`def a(allowed: String*) = P.within(allowed)`
creates `Seq[String] => Boolean` and not `String => Boolean` as the user
may intend


hubertp-test (approved):

hubertp (approved):

hubertp:
Sorry about the other approval, by mistake I ended up using my test account

PR #157, opened 2020-09-09
Graph.edgeCount more efficient: NodeDb.outEdgeCount without instantiating edges

mpollmeier:
includes refactorings in separate commit

### DIFF
--- a/core/src/main/java/overflowdb/NodeDb.java
+++ b/core/src/main/java/overflowdb/NodeDb.java
@@ -35,7 +35,7 @@ public abstract class NodeDb implements Node {
   private Object[] adjacentNodesWithProperties = new Object[0];
 
   /* store the start offset and length into the above `adjacentNodesWithProperties` array in an interleaved manner,
-   * i.e. each outgoing edge type has two entries in this array. */
+   * i.e. each adjancent edge type has two entries in this array. */
   private PackedIntArray edgeOffsets;
 
   /**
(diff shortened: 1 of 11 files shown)


core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -35,7 +35,7 @@
   private Object[] adjacentNodesWithProperties = new Object[0];
 
   /* store the start offset and length into the above `adjacentNodesWithProperties` array in an interleaved manner,
-   * i.e. each outgoing edge type has two entries in this array. */
+   * i.e. each adjancent edge type has two entries in this array. */
```
hubertp, line 38:
```suggestion
   * i.e. each adjacent edge type has two entries in this array. */
```

core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -421,6 +421,26 @@ public void addEdgeSilent(String label, Node inNode, Map<String, Object> keyValu
...
+  protected int outEdgeCount() {
+    int outEdgeCount = 0;
+    for (String label : layoutInformation().allowedOutEdgeLabels()) {
+      int offsetPos = getPositionInEdgeOffsets(Direction.OUT, label);
+      if (offsetPos != -1) {
+        int start = startIndex(offsetPos);
+        int length = blockLength(offsetPos);
+        int strideSize = getStrideSize(label);
+        for (int i = start;
+             i < adjacentNodesWithProperties.length && i < start + length;
```
hubertp, line 433: compute `start + length` and `adjacentNodesWithProperties.length` only once?

hubertp (approved):
nits

core/src/main/java/overflowdb/NodeDb.java
mpollmeier: `adjacentNodesWithProperties` is an array, so `.length` doesn't compute anything, and JIT will likely even inline it

`start + length`: yup ok will do across the board in this file

PR #158, opened 2020-09-11
random additional steps (plus testing some we just inherit from scala collections)

mpollmeier:


traversal/src/main/scala/overflowdb/traversal/Traversal.scala
```diff
@@ -47,6 +49,9 @@ class Traversal[A](elements: IterableOnce[A])
   def cast[B]: Traversal[B] =
     mapElements(_.asInstanceOf[B])
 
+  def collectAll[B <: A: ClassTag]: Traversal[B] =
```
hubertp, line 52: since you are adding docs, this is missing one ;)

traversal/src/main/scala/overflowdb/traversal/Traversal.scala
```diff
@@ -202,9 +207,41 @@ class Traversal[A](elements: IterableOnce[A])
       }.getOrElse(Traversal.empty)
     }
 
+  /** aggregate all objects at this point into the given collection (side effect)
+   * e.g. `mutable.ArrayBuffer.empty[A]` */
```
hubertp, line 211: the example is slightly confusing. maybe you meant
```
val xs = mutable.ArrayBuffer.empty[A]
ys.aggregate(xs)
```
?

traversal/src/main/scala/overflowdb/traversal/Traversal.scala
```diff
@@ -202,9 +207,41 @@ class Traversal[A](elements: IterableOnce[A])
...
+  /** sort elements by the value of the given transformation function */
+  @Doc("sort elements by the value of the given transformation function")
+  def sortBy[B](f: A => B)(implicit ord: Ordering[B]): Traversal[A] =
+    Traversal(elements.to(Seq).sortBy(f))
+
   def enablePathTracking: PathAwareTraversal[A] =
     PathAwareTraversal.from(elements)
 
+  /** group elements and count how often they appear */
+  @Doc("group elements and count how often they appear")
```
hubertp, line 229: is this a common pattern? I don't see how one cannot use the expression directly themselves given that `groupCount` is public. sometimes less is more

hubertp (approved):
nits, but looks ok

traversal/src/main/scala/overflowdb/traversal/Traversal.scala
mpollmeier: I don't understand "use the expression directly themselves" - can you elaborate please? do you mean just having `groupCount(A=>B)` and let the user figure out to pass `identity`?

hubertp: yeap

mpollmeier: in that case i'd like to keep the `.groupCount` without args, yes it can be quite useful
http://tinkerpop.apache.org/docs/current/reference/#groupcount-step

traversal/src/main/scala/overflowdb/traversal/Traversal.scala
mpollmeier: true, updated

traversal/src/main/scala/overflowdb/traversal/Traversal.scala
mpollmeier: done, i've also added docs for more steps

PR #161, opened 2020-09-18
Downgrade org.reflections to 0.9.11 due to a bug

hubertp:
Apparently I'm able to invoke our codepropertygraph's console via a path
that triggers https://github.com/ronmamo/reflections/issues/273.
This is very unfortunate and due to the way the artifact is being
published I'm unable to downgrade directly in the project I use.
Sadly this dependency is propagated very far down the chain and it
impossible to untangle it.

The fix is in 0.9.13-SNAPSHOT (not yet released) so suggesting to
downgrade until it is officially out.


hubertp:
Replaced with https://github.com/ShiftLeftSecurity/overflowdb/pull/162

PR #162, opened 2020-09-18
Replace reflections package with reflections8

hubertp:
Apparently I'm able to invoke our codepropertygraph's console via a path
that triggers ronmamo/reflections#273.
This is very unfortunate and due to the way the artifact is being
published I'm unable to downgrade directly in the project I use.
Sadly this dependency is propagated very far down the chain and it
impossible to untangle it.

The fix is in `0.9.13-SNAPSHOT` (not yet released) so suggesting to
downgrade until it is officially out.

Note: I was going to just downgrade to `0.9.11` but looks like that
version is pretty ancient. Instead I followed the same suggestion as in
https://github.com/JanusGraph/janusgraph/pull/2029
and replaced `reflections` with `reflections8` until `0.9.13` is
released.

### DIFF
--- a/traversal/build.sbt
+++ b/traversal/build.sbt
@@ -2,7 +2,7 @@ name := "overflowdb-traversal"
 
 libraryDependencies ++= Seq(
   "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-  "org.reflections" % "reflections" % "0.9.12",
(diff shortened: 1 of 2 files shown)


mpollmeier (approved):
cc @fabsx00 the joys of reflection :(

PR #166, opened 2020-10-20
Fix memory model violation / race condition: Dirty-flag must be volatile

bbrehm:
If I read https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html right, then we must make that flag volatile.


mpollmeier (approved):

mpollmeier:
Thank you! There's a ton more thread-unsafety in this codebase btw :)

PR #169, opened 2020-10-26
add storage format version in case we make further format changes in future

mpollmeier:
these changes are quite rare (we kept the initial format for almost
two years), but if they happen it's good to fail early and let the user
know

### DIFF
--- a/core/src/main/java/overflowdb/storage/BackwardsCompatibilityError.java
+++ b/core/src/main/java/overflowdb/storage/BackwardsCompatibilityError.java
@@ -1,0 +1,7 @@
+package overflowdb.storage;
+
+public class BackwardsCompatibilityError extends RuntimeException {
+  public BackwardsCompatibilityError(String msg) {
+    super(msg);
+  }
+}
--- a/core/src/main/java/overflowdb/storage/OdbStorage.java
+++ b/core/src/main/java/overflowdb/storage/OdbStorage.java
@@ -63,7 +63,7 @@ public class OdbStorage implements AutoCloseable {
     if (mvstoreFileMaybe.isPresent()) {
       this.doPersist = true;
       mvstoreFile = mvstoreFileMaybe.get();
-      if (mvstoreFile.exists()) {
+      if (mvstoreFile.exists() && mvstoreFile.length() > 0) {
         verifyStorageVersion();
       }
     } else {
--- a/core/src/main/java/overflowdb/storage/OdbStorage.java
+++ b/core/src/main/java/overflowdb/storage/OdbStorage.java
@@ -15,6 +15,10 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 public class OdbStorage implements AutoCloseable {
+  /** when persistence format changes, increase this number - this protects us from attempting to open outdated formats */
+  public static final int STORAGE_FORMAT_VERSION = 1;
+  public static final String METADATA_KEY_STORAGE_FORMAT_VERSION = "STORAGE_FORMAT_VERSION";
+
   private static final String INDEX_PREFIX = "index_";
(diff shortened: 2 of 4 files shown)


bbrehm:
It would be useful to have major/minor number, with semantics: If the file has unsupported major version, then give up. That way we would increase major number whenever we change format in a known-incompatible way, and increase minor number for supposedly-compatible changes. A way of doing that would be to set the minor version to the git-tag of the release.

In other words, the minor version number is insurance against fucking up compat: it helps us to make a new emergency release that can recover customer submissions that were generated with a buggy version.

We also need public API for accessing the version fields.

Furthermore, it would be nice if we could make the version field accessible _before_ attempting to open MVStore. However, I don't see a good place for that (after reading https://www.h2database.com/html/mvstore.html and inspecting a couple hexdumps). The proper place would be in the headers -- but that would require us to modify/extend MVStore, so it is probably not worth it.

Furthermore, we need additional version fields for the database schema, that are set on generation of a CPG. Again, major version (incremented whenever we do something known-incompatible -- not just database changes, but also changes in the meaning of cpg elements) and build versions (git tags of all relevant creators: odb, codepropertygraph, js2cpg, codescience, ocular, fuzzyc2cpg, etc). 

These fields are the responsibility of higher layers. But odb should offer an API for extracting version fields _before_ trying to read the nodes.. Let's make that one a follow-up PR, though.

mpollmeier:
Re major/minor versioning for the odb storage format: we can do all that, but given that we rarely change the schema at all (the last change was 18 months ago) I'm not convinced that it's overly useful. So far we chose the "pragmatic" route of breaking things in this very rare case, in order to avoid the "giant switch statement". We did so because we had everything under control and things were changing and breaking left and right - maybe that's not the case any more though. @fabsx00 thoughts?

Adding capabilities for versioning the application schema makes sense, I've had it on my radar for a while but didn't have an immediate need so far.

fabsx00 (approved):

bbrehm (changes requested):
Adding the minor version costs us nothing and will help us a lot -- at the very least on debugging.

I do not like giant switch statements either, but I 100% guarantee that our software will ship bugs (as all software does).

If we have an exact (minor) version, then we can debug and can recover customer artifacts during POCs. If we don't have exact versions, then we are thoroughly screwed.

This is not hypothetical: We had exactly this situation in the last 7 days, with a failure of forward compat. With an exact/minor version field, we can properly debug that, and write an ugly switch statement. Without a version field, we can basically gave up and wrote off the customer artifacts.

Switch statements that we don't write are not ugly. But if we need them, then they are probably less ugly than e.g. distributing one bazillion different versions of the docker image and trying them until one of them succeeds on the customer artifact. Even if we decide that one bazillion docker images are cleaner, let's cross that bridge when we get to it -- but by refusing a minor/exact version field, you lock in this decision now, with no justification.

mpollmeier:
We do have an 'exact' version - it's just not split into major/minor. But yes, splitting that up doesn't 'cost' much - will do.

bbrehm:
Where is the exact version? In that case, I withdraw my objections -- then we can just add documentation for that.

mpollmeier:
```
  /** when persistence format changes, increase this number - this protects us from attempting to open outdated formats */
  public static final int STORAGE_FORMAT_VERSION = 1;
```
https://github.com/ShiftLeftSecurity/overflowdb/pull/169/files#diff-f49645c084c26457a547e1fdb87bafc1edd7121841df1953623dad672b738217R19

mpollmeier:
It's EOD for you - I pushed my WIP status of splitting the version number into major/minor into https://github.com/ShiftLeftSecurity/overflowdb/pull/172 - let's continue the discussion there. 
I'll merge this one as is now, to unblock this and the two follow-up PRs.

PR #170, opened 2020-10-27
write int not strings for compactness, and some refactorings

mpollmeier:
* move (de)serializer out of storage for simplicity, adapt init order
* treat empty and non-existent storage files the same

note: contains https://github.com/ShiftLeftSecurity/overflowdb/pull/169 - review/merge that one first

### DIFF
--- a/core/src/main/java/overflowdb/storage/NodeDeserializer.java
+++ b/core/src/main/java/overflowdb/storage/NodeDeserializer.java
@@ -139,21 +139,6 @@ public class NodeDeserializer extends BookKeeper {
     }
   }
 
-  protected final Object[] toKeyValueArray(Map<String, Object> properties) {
-    List keyValues = new ArrayList(properties.size() * 2); // may grow bigger if there's list entries
-    for (Map.Entry<String, Object> entry : properties.entrySet()) {
-      //todo: We fail to properly intern strings contained in a List.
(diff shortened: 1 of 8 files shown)


fabsx00 (approved):

core/src/main/java/overflowdb/storage/NodeDeserializer.java
```diff
@@ -88,7 +91,8 @@ public final NodeRef deserializeRef(byte[] bytes) throws IOException {
     Object[] res = new Object[propertyCount * 2];
     int resIdx = 0;
     for (int propertyIdx = 0; propertyIdx < propertyCount; propertyIdx++) {
-      final String key = intern(unpacker.unpackString());
+      int keyId = unpacker.unpackInt();
+      final String key = intern(storage.reverseLookupStringToIntMapping(keyId));
```
bbrehm, line 95: I think that `storage.reverseLookupStringToIntMapping(keyId)` should already intern -- no need to look up in the hashmap all the time.

mpollmeier: I agree, but there's a few other issues with the interner, so I'll bring them in via another PR to not mix things too much.

mpollmeier:
two hours later... 
image

PR #172, opened 2020-10-27
WIP - split version into major/minor

mpollmeier:


bbrehm:
Ok, so this is not what I was asking for. Apologies for my unclear formulation.

I did not want a major/minor version split. I want an additional second version field that contains the commit-id of overflowdb (preferably in form of the git tag), as well as commit-ids / git-tags of all other involved projects. 

1. Backwards compatibility: Writer (e.g. js2cpg) is newer than reader (e.g. codescience / ocular / inspect): In this case, your original proposal with an unsplit STORAGE_VERSION field is fine. No objections: We give a good, i.e. actionable, error message instead of silently failing or computing wrong results. The required action is "update codescience / ocular / inspect". This should never happen in our cloud deployment: We roll out the new backend before distributing the new frontend to users.

2. Forwards compatibility: Reader (e.g. codescience / ocular / inspect) is newer than writer (e.g. js2cpg): Here, I see a problem. The issue is: The trivial action on error (update the frontend, e.g. js2cpg) is unavailable -- often we are in the situation that we have a customer artifact, generated with some version of the frontend, and need to analyze it. We don't have the code for the customer project, so we cannot regenerate it -- that requires manual intervention on the side of our users. In other words, in a good solution, we must be able to deal with this.

2.1. Planned update: Suppose we want to change the storage format. We can 

2.1.1. give up on compat: We update our cloud deployment, and refuse to handle old odbs. This means that we impose an effective downtime on all our customer: the critical window between sever-side-update deployment and client-side-update. By scheduling this to ungodly hours, and announcing it beforehand, and keeping the window short, we may get away with that. I don't know what kind of SLAs we usually sign.

2.1.2. For the critical window, we have two versions deployed, and select the right one for the customer submission. Infra will hate us for this.

2.1.3. When changing the storage format, we always keep compat code: That is, instead of changing the code, we clone it into a new deserializer and keep the old one alive for the duration of the critical window. Then, on loading, we select the right one.

3. Dealing with bugs. Suppose we have a bug. This can lead to unintentional forwards compatibility failures: Newer versions of overflowdb / backend cannot correctly read odbs written by an older version of overflowdb / frontend. We have tests, but sometimes the stars will align and something slips through. Now there are two possibilities:

3.1. We only have the version field you proposed. This version field was not incremented, because it was not our intent to break compatibility. We now get a lot of analysis failures, and we don't know the reasons for that: Are these unrelated problems, or are these all the same issue? So we now need to manually debug a ton of failures. Once we debugged these failures, we may need to ask some customers for resubmissions. Or we may be able to deploy a new patched overflowdb version in the backend. This new patched backend version would need to be able to deal with databases emitted by three versions: (a) pre-unintentional-breakage, (b) during unintentional-breackage, (c) post-unintentional-breakage. It would need to deal with all three different versions *without knowing the unintentional storage version of the artifact*. If we can't pull that off, then we have caused effective downtime and require manual intervention by our users.

3.2. We also have the precise version (git tag or commit) of the writer / frontend, including precise version of their overflowdb (dynamically linked! Doesn't matter what we write in some `build.sbt`!). Now it becomes much, much easier to triage issues caused by this bug: Just look at the logs to see the versions. It also becomes feasible to deal with this bug without inconveniencing users: Now, upon reading, we _know_ whether the database is written in the effective formats (a), (b), or (c) -- so we may be able to write a simple switch statement to deal with the differences.


In other words: I want the precise versions embedded in the mvstore (not something that has been manually incremented because somebody intentionally broke stuff; instead, the precise actual version that identifies what actual classfiles we're running). This simplifies debugging, and is insurance against unintentional fuckups. The scenario 3.1. is not hypothetical, it happened last week. 

An ugly (and temporary!) switch statement would have been preferable to the actual course of action (time wasted triaging duplicates of the issue, inconveniencing customers by requesting manual regeneration of artifacts -- i.e. effective multi-day downtime for all affected customer projects).

cc @hubertp @ml86 @fabsx00 for comments. 

If you don't understand what I'm writing, please ask me to clarify.

mpollmeier:
Thanks for clarifying - here's a few thoughts:

Getting the version number is easy: `classOf[overflowdb.Graph].getPackage.getImplementationVersion`. 
When opening a new graph we could pass 'relevant classes' in the config and simply persist their version numbers in the metadata table. 
When opening an existing storage we can easily provide an api to compare the current runtime versions with the versions this storage was created with. 
So far so good. 

However: what if a storage was opened/closed multiple times by different applications that have differnt versions of the same dependencies? E.g. the initial creator (frontend) uses odb v1, then it's adapted by an intermediate process (e.g. cpg2sp) using odb v3, and finally it's opened by something (e.g. (sp2x) using odb v2. How would you like to handle this situation? 

Note1: this is simplified, the other dependencies obviously also come into the mix.  
Note2: nodes that aren't read/changed will not be serialized again, so if there's incompatible changes between the dependencies, it will only blow up once the nodes are touched.

bbrehm:
>However: what if a storage was opened/closed multiple times by different applications that have differnt versions of the same dependencies? E.g. the initial creator (frontend) uses odb v1, then it's adapted by an intermediate process (e.g. cpg2sp) using odb v3, and finally it's opened by something (e.g. (sp2x) using odb v2. How would you like to handle this situation?

I'd tend to persist a lot of info: The reason for the field is unknown future bugs, and we don't know exactly what info we will need. So, version-info could be `List/Array[Map[FullyQualifiedClassName, ImplementationVersion]]` -- and a new Map gets added on opening an overflowdb.

In addition to API for extracting this version-info (e.g. just return the list), we also need some `register_version(class)` API function, that adds a class to the current (i.e. last) version-map. This is because the cpg-loader doesn't know yet which classes are relevant. So e.g. `securityprofile` or `semanticcpg` could begin by registering themselves; and odb could begin by registering itself.

Regarding what to actually do about the versions (besides persisting and logging it), I'm happy to cross that bridge when we reach it: Initially we do nothing, and if we push incompatible version changes, then the most likely action will be to rewrite the entire database into the new format (everything else is probably unmaintainable; this action is slowish, but needs to be done only once).

Realistically, we deal with two bundles of versions: The frontend version that created the cpg, and the backend version that computes enhancements.

mpollmeier:
https://github.com/ShiftLeftSecurity/overflowdb/pull/176

PR #176, opened 2020-11-03
library version metadata to enable version compatibility handling

mpollmeier:
* allows to register random versions for any potentially relevant library
* persists new set of versions across restarts
* automatically persists overflowdb version
* re https://github.com/ShiftLeftSecurity/overflowdb/pull/172#issuecomment-720324614

### DIFF
--- a/core/src/main/java/overflowdb/Graph.java
+++ b/core/src/main/java/overflowdb/Graph.java
@@ -12,8 +12,8 @@ import overflowdb.util.PropertyHelper;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


bbrehm:
Very good. The API looks mostly fine -- but why `Collection<Map<String, String>>`? We want to guarantee fast stable access by integer index.

Why the complicated thing with string parsing? Can't we store something like a simple `ArrayList[Map[String, String]]`?

mpollmeier:
i changed it to ArrayList

Mvstore gives us (multiple) `Map<A, B>` where A and B need to be (de-)serializable. in theory we could add multiple maps to avoid this fun string parsing, but having many maps led to problems in the past, and I'd like to keep us as independent of the underlying storage as possible. 
The 'cleaner' alternative would be to have another binary format just like we do for nodes with MsgPack. 
http://www.h2database.com/html/mvstore.html#maps

bbrehm (approved):

PR #178, opened 2020-11-16
use github actions instead of travis-ci.org

mpollmeier:


mpollmeier:
ready for review, but only merge once secrets have been configured

.github/workflows/pr.yml
```diff
@@ -0,0 +1,17 @@
+name: PR
+on: pull_request
+jobs:
+  pr:
+    runs-on: ubuntu-18.04
```
hubertp, line 5:
```suggestion
    runs-on: ubuntu-20.04
```
Let's stick to something a bit more recent?

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
+name: Release
+on:
+  push:
+    branches: [master, main]
+    tags: ["*"]
+jobs:
+  release:
+    runs-on: ubuntu-18.04
```
hubertp, line 8:
```suggestion
    runs-on: ubuntu-20.04
```
same reason as before

hubertp (approved):
nits

fabsx00 (approved):

.github/workflows/pr.yml
mpollmeier: From https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
image

It probably wouldn't harm, but..

mpollmeier: discussed via slack, we'll stay with 18.04 for now

PR #184, opened 2021-01-23
next/nextOption should have parens since they are side effecty

mpollmeier:
note: same goes for `.l` etc but since it's part of the core dsl,
conciseness wins over consistency in those cases


mpollmeier:
@max-leuthaeuser wait with the review for a moment, there's a few more similar changes coming

mpollmeier:
@max-leuthaeuser good to go :)

max-leuthaeuser:
> note: same goes for `.l` etc but since it's part of the core dsl,
> conciseness wins over consistency in those cases

I am totally with you at this point 👍

max-leuthaeuser (approved):

PR #194, opened 2021-06-01
improve speed of NodeRef hashing

bbrehm:
Found this when reading profiling data. This gives about 5% speedup on dataflow-heavy loads.

### DIFF
--- a/core/src/main/java/overflowdb/NodeDb.java
+++ b/core/src/main/java/overflowdb/NodeDb.java
@@ -754,12 +754,12 @@ public abstract class NodeDb implements Node {
      * The style (shift-xor 33 and multiply) is similar to murmur3; the multiply constant is randomly chosen odd number.
      * Feel free to change this.
      * */
-    long tmp = (id() ^ (id() >>> 33)) * 0x1ca213a8d7b7d9b1L;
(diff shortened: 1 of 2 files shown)


mpollmeier (approved):
When you say 5% faster, does this refer to one of our applications like running cpg2sp, or do you mean just the hashing function is 5% faster this way?

Can you please add a comment to explain why we do this and what it roughly does?

ml86:
I guess you did not come up with the shift pattern and the used constant on your own. Can you please put a comment which documents where this comes from.

bbrehm:
The thing that gets 5% speedup are some end-to-end cpg2sp runs.

I think this is because the JIT failed to properly inline / optimize away the vararg calling convention?

The shift-and-multiply pattern comes from murmur3; the constant multiplier is chosen by a fair die (`randLong() | 1`). This is not the same as actual murmur mixing function (that uses two multiplies).

We don't need a good hash here, only something fast; but we can spare the cycle to be at least better than the java built-in hash function (afaiu `int hashCode(long x){return (int) (x ^ (x>>>32));}

bbrehm:
@mpollmeier do we want NodeRef and NodeDb with the same id to compare equal?

In the old code, `someRef.equals(someDb) == true`, but `someDb.equals(someRef) == false`.

mpollmeier:
no i'd say they should not be equal

PR #197, opened 2021-07-04
mutable set is more performant than immutable set...

mpollmeier:


mpollmeier:
re https://github.com/ShiftLeftSecurity/codescience/pull/5100

ml86 (approved):

PR #202, opened 2021-07-21
msgpack doesn't support double, so we shouldn't do that either for now

mpollmeier:


bbrehm:
What about https://www.javadoc.io/doc/org.msgpack/msgpack-core/0.8.7/org/msgpack/core/MessagePacker.html#packDouble-double- ?

mpollmeier:
ouch, not sure how i managed to miss that? thank you!

mpollmeier:
oh, i think i know how i managed to miss this early: the reader api is a bit confusing: it does support double, but sounds like it only supports float: `value.asFloatValue().toDouble();`
will send a follow up pr shortly

PR #204, opened 2021-07-26
allow to convert properties for storage

mpollmeier:

PR #206, opened 2021-07-31
optional conversion of property values prior to serialization

mpollmeier:
* supersedes https://github.com/ShiftLeftSecurity/overflowdb/pull/204 with a nicer design: rather than providing the conversion for node and edge properties on NodeDb, we can now register a function when opening the graph - the conversion is always the same anyway
* also: support for primitive arrays without boxing

related: https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/91


bbrehm (approved):
LGTM

PR #208, opened 2021-08-05
also use IndexedSeq in path

mpollmeier:
* whilst we're at it and using IndexedSeq for list properties
* Vector was used before and that's an IndexedSeq as well, so we're not
  really changing much besides using a more general api

### DIFF
--- a/traversal/src/main/scala/overflowdb/traversal/PathAwareRepeatStep.scala
+++ b/traversal/src/main/scala/overflowdb/traversal/PathAwareRepeatStep.scala
@@ -19,9 +19,9 @@ object PathAwareRepeatStep {
       case SearchAlgorithm.BreadthFirst => new FifoWorklist[A]
     }
 
-    element: A => new PathAwareTraversal[A](new Iterator[(A, Vector[Any])] {
+    element: A => new PathAwareTraversal[A](new Iterator[(A, IndexedSeq[Any])] {
       val visited = mutable.Set.empty[A]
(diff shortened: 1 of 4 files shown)


bbrehm:
I'm not so sure.

As you noted, Vector is more specific than IndexedSeq, and return type is covariant, not contravariant, so the API change is technically breaking.

The actual types used are not changed by this (`Vector.empty` is the same as `IndexedSeq.empty`), only the declared return type.

However, this may makes the code harder to read, e.g. https://github.com/ShiftLeftSecurity/overflowdb/blob/58f0804b68dd44dc0a80660f7ed8d95a6265a43c/traversal/src/main/scala/overflowdb/traversal/PathAwareTraversal.scala#L15 

`IndexedSeq.appended` is N(log N) for `Vector` and instances share storage (i.e. two long paths branching off share the memory allocations for the initial piece). For `ArraySeq` appending is O(N) and no memory is shared, leading to O(N^2) cost for constructing a path of length N.

Hence, calling `.appended` on `IndexedSeq` is an immediate red flag -- if we go through with this change, we should definitely add source code comments that these are really `Vector`, to simplify the job of understanding the code for "future us".

mpollmeier:
I'm not so afraid of breaking api changes at this point (esp in this rarely used area), but the performance considerations make sense. 
I just thought it's a nice consolidation, but probably not worth it.

bbrehm:
There are no performance impacts of this PR -- just readability considerations: the old code has obviously the right complexity class, and the new code has non-obviously the right complexity class. So we could change this if there is a good reason, and then add lots of source-code comments to replace the lost readability.

PR #209, opened 2021-08-08
fix race conditions between writes and reads of adjacent nodes/edges/edgeProperties

mpollmeier:
* use writer mutex to synchronize all write operations
* use container for adjacentNodesWithEdgeProperties / edgeOffsets
* make container volatile to prevent bugs via cpu/compiler optimizations such as statement reordering
* container only needs to be recreated on `grow` and `trim` - other update operations may result in reading outdated data, but ensures better read performance
* DummyEdgeIterator: handle null case in array: hold 'peeked' entry in class: underlying array may have changed (e.g. if an edge was removed concurrently). We want to at least guarantee correct Iterator semantics, i.e. if `hasNext` was true, then `.next` should never be null or throw an exception
* related: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1319#issuecomment-894121966

### DIFF
--- a/core/src/main/java/overflowdb/AdjacentNodes.java
+++ b/core/src/main/java/overflowdb/AdjacentNodes.java
@@ -1,0 +1,29 @@
+package overflowdb;
+
+import overflowdb.util.PackedIntArray;
+
+public class AdjacentNodes {
+
+  /**
+   * holds refs to all adjacent nodes (a.k.a. dummy edges) and the edge properties
+   */
+  public final Object[] nodesWithEdgeProperties;
+
+  /* store the start offset and length into the above `adjacentNodesWithEdgeProperties` array in an interleaved manner,
+   * i.e. each adjacent edge type has two entries in this array. */
+  public final PackedIntArray edgeOffsets;
+
+  /**
+   * create an empty AdjacentNodes container for a node with @param numberOfDifferentAdjacentTypes
+   */
+  public AdjacentNodes(int numberOfDifferentAdjacentTypes) {
+    nodesWithEdgeProperties = new Object[0];
+    edgeOffsets = PackedIntArray.create(numberOfDifferentAdjacentTypes * 2);
+  }
+
+  public AdjacentNodes(Object[] nodesWithEdgeProperties, PackedIntArray edgeOffsets) {
+    this.nodesWithEdgeProperties = nodesWithEdgeProperties;
+    this.edgeOffsets = edgeOffsets;
+  }
+
+}
--- a/core/src/main/java/overflowdb/NodeDb.java
+++ b/core/src/main/java/overflowdb/NodeDb.java
@@ -236,12 +236,12 @@ public abstract class NodeDb extends Node {
                             Edge edge,
                             int blockOffset,
                             String key) {
-    AdjacentNodes adjacentNodes = this.adjacentNodes;
-    int propertyPosition = getEdgePropertyIndex(adjacentNodes, direction, edge.label(), key, blockOffset);
+    AdjacentNodes adjacentNodesTmp = this.adjacentNodes;
+    int propertyPosition = getEdgePropertyIndex(adjacentNodesTmp, direction, edge.label(), key, blockOffset);
     if (propertyPosition == -1) {
       return null;
     }
-    return (P) adjacentNodes.nodesWithEdgeProperties[propertyPosition];
+    return (P) adjacentNodesTmp.nodesWithEdgeProperties[propertyPosition];
   }
 
   public <V> void setEdgeProperty(Direction direction,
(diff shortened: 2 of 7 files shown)


core/src/main/java/overflowdb/AdjacentNodes.java
```diff
@@ -0,0 +1,29 @@
...
+public class AdjacentNodes {
+
+  /**
+   * holds refs to all adjacent nodes (a.k.a. dummy edges) and the edge properties
+   */
+  public final Object[] nodesWithEdgeProperties;
+
+  /* store the start offset and length into the above `adjacentNodesWithEdgeProperties` array in an interleaved manner,
+   * i.e. each adjacent edge type has two entries in this array. */
+  public final PackedIntArray edgeOffsets;
```
bbrehm, line 14: I think we should inline the PackedIntArray here: That's otherwise one extra indirection that is unlikely to be in cache, and main memory is far far away.

core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -12,9 +12,9 @@
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.locks.ReentrantLock;
```
bbrehm, line 17: I don't think that we need an extra ReentrantLock: `synchronized` / the monitor does the job just fine.

core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -248,57 +236,60 @@ public void markAsClean() {
                             Edge edge,
                             int blockOffset,
                             String key) {
-    int propertyPosition = getEdgePropertyIndex(direction, edge.label(), key, blockOffset);
+    AdjacentNodes adjacentNodes = this.adjacentNodes;
```
bbrehm, line 239: This does the right thing, but can we choose a different name for the temp? Just for readability -- make it obvious that we do the right thing, and make it easy to grep for accesses to the instance variable in order to hunt for violations.

core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -248,57 +236,60 @@ public void markAsClean() {
...
   public <V> void setEdgeProperty(Direction direction,
                                   String edgeLabel,
                                   String key,
                                   V value,
                                   int blockOffset) {
-    int propertyPosition = getEdgePropertyIndex(direction, edgeLabel, key, blockOffset);
-    if (propertyPosition == -1) {
-      throw new RuntimeException("Edge " + edgeLabel + " does not support property `" + key + "`.");
+    try {
+      writerMutex.lock();
```
bbrehm, line 253: This can just be `synchronized` on the NodeDb instance.

core/src/main/java/overflowdb/util/DummyEdgeIterator.java
```diff
@@ -17,6 +17,9 @@
   private final String label;
   private final NodeRef thisRef;
 
+  /** used for peeking */
+  private Edge nextCached;
+
   public DummyEdgeIterator(Object[] array, int begin, int exclusiveEnd, int strideSize,
                            Direction direction, String label, NodeRef thisRef) {
```
bbrehm, line 24: Let's document that this is a benign race condition (`array` can be modified under us, but this is OK and we spent time thinking about the fact).

bbrehm (changes requested):

core/src/main/java/overflowdb/NodeDb.java
mpollmeier: ah, you mean just `synchronized(this)`? yeah true, that will do the the same and is more concise :+1:

core/src/main/java/overflowdb/NodeDb.java
mpollmeier: sure, no worries

mpollmeier: just renamed those ones to `adjacentNodesTmp`

core/src/main/java/overflowdb/NodeDb.java
mpollmeier: actually, does it really? we want all write operations to be synchronized on the same lock, i.e. if one thread is setting an edge property we don't want to grow the adjacentNodes array at the same time in a different thread...

core/src/main/java/overflowdb/util/DummyEdgeIterator.java
mpollmeier: done

core/src/main/java/overflowdb/AdjacentNodes.java
mpollmeier: as discussed, via follow-up pr

core/src/main/java/overflowdb/NodeDb.java
mpollmeier: as discussed: will push as synchronized because all synchronized(this) blocks lock on the same object monitor

bbrehm (approved):
Looks good!

PR #210, opened 2021-08-10
unwrap packed array

bbrehm:
This unwraps the PackedIntArray, in order to reduce the number of memory hops for reading neighbors/edges by one, i.e. restore the status from before fixing the race.


mpollmeier:
easiest way to resolve the conflicts is probably to cherry pick your commit onto master... my PR was squashed on merge

mpollmeier:
I guess it's time to bring in an (automatic) code formatter? wdyt @ml86? 
Is there a good tool for java? 
I don't want the same situation as in java2cpg where the formatting tool tells us to e.g. order imports alphabetically - I want the formatter to just do that for us.

ml86 (approved):

core/src/main/java/overflowdb/AdjacentNodes.java
```diff
@@ -10,20 +8,86 @@
...
+    if(offsets instanceof byte[]){
+      return ((byte[]) offsets)[pos];
+    } else if(offsets instanceof short[]){
+      return ((short[]) offsets)[pos];
+    } else if (offsets instanceof int[]){
+      return ((int[]) offsets)[pos];
+    } else throw new RuntimeException(offsets.getClass().getName());
+  }
+
+  //returns null if the insertion was in-place; returns a new AdjacentNodes if we need to update
```
mpollmeier, line 42: i really don't like returning `null` as it's setting us up for shooting us in the foot, especially in a `public` method...
1) you could probably make it package-private, i.e. just drop `public`?
2) i may have missed something, but can't we just always `return offsets` here? I didn't go through the callsites in detail but it looked like it wouldn't break them..?

core/src/main/java/overflowdb/NodeDb.java
```diff
@@ -637,7 +637,7 @@ private final synchronized int storeAdjacentNode(Direction direction, String edg
     int strideSize = getStrideSize(edgeLabel);
 
     Object[] adjacentNodesWithEdgeProperties = adjacentNodes.nodesWithEdgeProperties;
-    int edgeOffsetLengthB2 = adjacentNodes.edgeOffsets.length() >> 1;
+    int edgeOffsetLengthB2 = adjacentNodes.getIntOffsetLen() >> 1;
```
mpollmeier, line 640: Hmm, I thought `edgeOffsets.length` was more helpful to the reader than `getIntOffsetLen`. 
Not a very clear winner here I guess, maybe something like `adjacentNodes.offsetLength()`?

mpollmeier (commented):
Didn't we discuss to keep the code for PackedIntArray in that file, and only change the style from java-style virtual dispatch `edgeOffset.get(x)` to c-style static invoke `PackedIntArray.get(edgeOffset, x)` for performance? 

This PR completely inlines the code, which means we can't reuse PackedIntArray any more and IMO it's a little harder to read because AdjacentNodes now comprises a few different things. 

If there's a reason to do it this way that's ok, I'm just wondering why you decided to change course, or if I misunderstood something in our call.

core/src/main/java/overflowdb/NodeDb.java
bbrehm: `edgeOffsets.length` was more helpful to the reader. However, `edgeOffsets` as an object doesn't exist anymore, so we can't ask it for its length.

But I'm happy to change the name, if you have any suggestions!

core/src/main/java/overflowdb/AdjacentNodes.java
bbrehm: :+1: to making everything package private!

core/src/main/java/overflowdb/NodeDb.java
bbrehm: Ok, renamed it

core/src/main/java/overflowdb/AdjacentNodes.java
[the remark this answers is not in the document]
bbrehm: Made it package-private

core/src/main/java/overflowdb/AdjacentNodes.java
```diff
@@ -10,20 +8,86 @@
...
 
-  public AdjacentNodes(Object[] nodesWithEdgeProperties, PackedIntArray edgeOffsets) {
+
+  public AdjacentNodes(Object[] nodesWithEdgeProperties, Object offsets) {
     this.nodesWithEdgeProperties = nodesWithEdgeProperties;
-    this.edgeOffsets = edgeOffsets;
+    if(offsets == null) throw new RuntimeException("NPE");
+    if(offsets instanceof byte[] | offsets instanceof short[] | offsets instanceof int[])
+      this.offsets = offsets;
+    else throw new RuntimeException();
```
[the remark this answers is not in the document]
bbrehm: Removed the check -- the constructor is package private now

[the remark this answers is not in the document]
mpollmeier: sorry I wasn't very precise here - what i actually meant was to add a message to the exception - I'll send a proposal

core/src/main/java/overflowdb/AdjacentNodes.java
```diff
@@ -1,29 +1,99 @@
...
+   * getOffset(2 * kindOffset + 1) gets the length of the relevant slice in nodesWithEdgeProperties of the desired kindOffset (get the kindOffset from layout info)
+   */
+  public int getOffset(int pos){
+    if(offsets instanceof byte[]){
+      return ((byte[]) offsets)[pos];
+    } else if(offsets instanceof short[]){
+      return ((short[]) offsets)[pos];
+    } else if (offsets instanceof int[]){
+      return ((int[]) offsets)[pos];
+    } else throw new RuntimeException();
```
mpollmeier, line 44:
```suggestion
    } else throw new RuntimeException("corrupt state: offsets of type " + offsets.getClass);
```

core/src/main/java/overflowdb/AdjacentNodes.java
```diff
@@ -1,29 +1,99 @@
...
+          newOffsets[i] = oldOffsets[i];
+        }
+        newOffsets[pos] = val;
+        return new AdjacentNodes(nodesWithEdgeProperties, newOffsets);
+      }
+    } else if (offsets instanceof int[]){
+      ((int[]) offsets)[pos] = val;
+      return this;
+    } else {
+      throw new RuntimeException();
```
mpollmeier, line 88:
```suggestion
      throw new RuntimeException("corrupt state: offsets of type " + offsets.getClass);
```

core/src/main/java/overflowdb/AdjacentNodes.java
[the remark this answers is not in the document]
mpollmeier: :+1:

mpollmeier: :+1:

mpollmeier (approved):

PR #211, opened 2021-08-12
Delete temporary files as early as possible.

Ferada:
This means that there's no chance on *nix systems that the `/tmp` directory can end up with temporary files left over when the JVM gets killed.

At least on systems that aren't Windows, since there we can't simply delete the still open file AFAIK.

This would be cleaner if done via `StandardOpenOption.DELETE_ON_CLOSE` on the part that opens the file (in `FileStore`), however even overloading wouldn't allow us to do that, instead the whole `FileStore` interface would have to be redone I think.

Anyway, just an idea, since it doesn't require too many changes.

### DIFF
--- a/core/src/main/java/overflowdb/storage/OdbStorage.java
+++ b/core/src/main/java/overflowdb/storage/OdbStorage.java
@@ -65,7 +65,12 @@ public class OdbStorage implements AutoCloseable {
         mvstoreFile = File.createTempFile("mvstore", ".bin");
         if (!System.getProperty("os.name").toLowerCase().contains("win")) {
           mvstoreFileStore = new FileStore();
-          mvstoreFileStore.open(mvstoreFile.getAbsolutePath(), false, null);
+          boolean readOnly = false;
+          char[] encryptionKey = null;
+          mvstoreFileStore.open(mvstoreFile.getAbsolutePath(), readOnly, encryptionKey);
+          /** Note: we're deleting the temporary storage file as early as possible on *nix systems, i.e. while it's still running
+            * This is so we don't fill up `/tmp` if the JVM gets killed.
+            **/
           mvstoreFile.delete();
         } else {
           mvstoreFile.deleteOnExit();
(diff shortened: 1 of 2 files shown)


core/src/main/java/overflowdb/storage/OdbStorage.java
```diff
@@ -54,17 +55,21 @@ public static OdbStorage createWithSpecificLocation(final File mvstoreFile) {
...
         initializeStringToIntMaxId();
       }
     } else {
       try {
-        this.doPersist = false;
         mvstoreFile = File.createTempFile("mvstore", ".bin");
-        mvstoreFile.deleteOnExit(); // `.close` will also delete it, this is just in case users forget to call it
+        if (!System.getProperty("os.name").toLowerCase().contains("win")) {
+          mvstoreFileStore = new FileStore();
+          mvstoreFileStore.open(mvstoreFile.getAbsolutePath(), false, null);
```
mpollmeier, line 68: better readability?

```suggestion
          boolean readOnly = false;
          char[] encryptionKey = null;
          mvstoreFileStore.open(mvstoreFile.getAbsolutePath(), readOnly, encryptionKey);
```

core/src/main/java/overflowdb/storage/OdbStorage.java
```diff
@@ -54,17 +55,21 @@ public static OdbStorage createWithSpecificLocation(final File mvstoreFile) {
...
+          mvstoreFile.delete();
```
mpollmeier, line 69:
```suggestion
          /** Note: we're deleting the temporary storage files as early as possible on *nix systems, i.e. while it's still running
            * This is so we don't fill up `/tmp` if the JVM get's killed. 
            **/
          mvstoreFile.delete();
```

mpollmeier (approved):
Rather than a semi-fishy workaround like this I'd prefer everybody would shut down their JVMs properly rather than killing them, but I guess you've already tried that route...?

Ferada:
> Rather than a semi-fishy workaround like this I'd prefer everybody would shut down their JVMs properly rather than killing them

True, but often you don't get the luxury. The pattern's quite common though - there's simply no other way of guaranteeing that temporary files get cleaned up when a process can be killed at any time.

PR #216, opened 2021-08-31
don't use ReferenceManager if overflow is disabled

mpollmeier:
For now this is mostly a cleanup/refactoring exercise, but we'll also save a bit of memory by not holding the 'clearable refs' queue in ReferenceManager. More importantly, this is a prerequisite for a different optimisation that I'm working on.

### DIFF
--- a/core/src/main/java/overflowdb/NodeRef.java
+++ b/core/src/main/java/overflowdb/NodeRef.java
@@ -55,8 +55,8 @@ public abstract class NodeRef<N extends NodeDb> extends Node {
     return node == null;
   }
 
-  /* only called by @ReferenceManager */
-  protected void clear() {
+  /* only supposed to be called by @NodesWriter */
+  public void clear() {
     this.node = null;
   }
 
--- a/core/src/main/java/overflowdb/NodeRef.java
+++ b/core/src/main/java/overflowdb/NodeRef.java
@@ -101,7 +101,7 @@ public abstract class NodeRef<N extends NodeDb> extends Node {
       final N node = readFromDisk(id);
       if (node == null) throw new IllegalStateException("unable to read node from disk; id=" + id);
       this.node = node;
-      graph.referenceManager.registerRef(this); // so it can be cleared on low memory
+      graph.registerNodeRef(this);
       return node;
     }
   }
--- a/core/src/main/java/overflowdb/NodeRef.java
+++ b/core/src/main/java/overflowdb/NodeRef.java
@@ -55,9 +55,13 @@ public abstract class NodeRef<N extends NodeDb> extends Node {
     return node == null;
   }
 
-  /* only supposed to be called by @NodesWriter */
-  public void clear() {
-    this.node = null;
+  /**
+   * Only supposed to be called by @NodesWriter
+   * We'd prefer this to be package-private, but since NodesWriter is in a different package that's not an option in java.
+   * To not pollute the public api (esp. for console users) we made this method static instead.
+   * */
+  public static void clear(NodeRef ref) {
+    ref.node = null;
   }
 
   protected byte[] serializeWhenDirty() {
@@ -133,7 +137,7 @@ public abstract class NodeRef<N extends NodeDb> extends Node {
   @Override
   public void remove() {
     get().remove();
-    clear();
+    NodeRef.clear(this);
   }
 
   @Override
--- a/core/src/main/java/overflowdb/storage/NodesWriter.java
+++ b/core/src/main/java/overflowdb/storage/NodesWriter.java
@@ -47,7 +47,7 @@ public class NodesWriter {
         try {
           byte[] bytes = nodeSerializer.serialize(nodeDb);
           storage.persist(ref.id(), bytes);
-          ref.clear();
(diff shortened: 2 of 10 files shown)


core/src/main/java/overflowdb/NodeRef.java
```diff
@@ -55,8 +55,8 @@ public boolean isCleared() {
     return node == null;
   }
 
-  /* only called by @ReferenceManager */
-  protected void clear() {
+  /* only supposed to be called by @NodesWriter */
+  public void clear() {
```
bbrehm, line 59: Can we make this package-private? Or alternatively make it a static method (i.e. `static void clearNode(NodeRef ref)`)?

Point is to avoid cluttering the REPL completion in ocular/joern

core/src/main/java/overflowdb/storage/NodesWriter.java
```diff
@@ -0,0 +1,78 @@
...
+    this.storage = storage;
+  }
+
+  /**
+   * writes all references to storage, blocks until complete.
+   */
+  public void writeAndClearBatched(Spliterator<? extends Node> nodes, int estimatedTotalCount) {
+    CountEstimates countEstimates = new CountEstimates();
+
+    StreamSupport.stream(nodes, false).forEach(node -> {
```
bbrehm, line 35: Can we add a comment that this is sequential, and the reason that it has to be sequential? At least I had a hiccup because I first pattern-matched and assumed parallel stream, searching for races, until I thought to look up the docs for `java.util.stream.StreamSupport.stream`...

core/src/main/java/overflowdb/storage/NodesWriter.java
```diff
@@ -0,0 +1,78 @@
...
+          float progressPercent = 100f * countEstimates.countEstimate / estimatedTotalCount;
+          countEstimates.nodesToSerializeUntilLogging = LogAfterXSerializedNodes;
+          logger.info(String.format("progress of writing nodes to storage: %.2f%s", Float.min(100f, progressPercent), "%"));
+        }
+      }
+    });
+  }
+
+  /**
+   * Counts aren't atomic or synchronized, but worst case we'll miss or duplicate a log.info message,
```
bbrehm, line 69: This comment is misleading -- the counts don't need to be atomic or synchronized because they are not accessed concurrently.

If they were writtenn concurrently, in a relatively tight loop as above, then I'd advise to change something: We might know that the race is benign, but the CPU doesn't. More than one core accessing the same cacheline with at least one writer is always slow (correctness depends on atomic/volatile/etc, whether we access the same / overlapping addresses, and application logic; perf is always bad).

bbrehm (approved):
Minor nits -- if you disagree with them, then you can merge without addressing.

core/src/main/java/overflowdb/NodeRef.java
mpollmeier: Good point, we don't really want users to see this. 
`NodesWriter`is in the `storage` subpackage.... I'm debating with myself if that should move or we should go for your static method suggestion.... I think I prefer the latter

core/src/main/java/overflowdb/storage/NodesWriter.java
mpollmeier: Actually, I just went through the git history and saw that Denis made the serialization parallel, but made the 'write to storage' sequential. I think he measured the MVStore performance, and given that we're writing to disk, that probably makes sense. Unfortunately he didn't add comments (besides the commit title), but I'll change that. 

Thanks for bringing this up, it triggered me to dig deeper.

core/src/main/java/overflowdb/storage/NodesWriter.java
mpollmeier: I was using a parallel stream when I wrote that, and then had to change it to sequential due to race conditions. As noted above we'll go back to semi-parallel, which makes the comment accurate again I suppose. 

I don't mind so much about correctness here, but performance is important. 
My first naive thought was to batch the parallel streams, i.e. partition the nodes into chunks of e.g. 1024 and then use the same 'shared roughly correct racy counter'. That way we can scale down the "performance is bad" aspect of this. 
Do you have any suggestions?

core/src/main/java/overflowdb/NodeRef.java
mpollmeier: done in 3ff21af

core/src/main/java/overflowdb/storage/NodesWriter.java
mpollmeier: historical reference: https://github.com/ShiftLeftSecurity/overflowdb/pull/67

core/src/main/java/overflowdb/storage/NodesWriter.java
bbrehm: If we partition the node into chunks, then we can get the best of all worlds: Use an atomic counter (so no races), but only increment once a chunk has finished.

Alternatively we can use an atomic and assume that serialization+mvstore is so expensive that the (contended) atomic doesn't matter: Now that I think about it, this probably is the case.

PR #218, opened 2021-09-07
build on jdk8 again

mpollmeier:
We never wanted to stop targeting jdk8, and the recent upgrade to
building on jdk11 while keeping the source/target params to 1.8 did
not work. Buildiung on jdk8 again is the safest bet for now.


ursachec (approved):

mpollmeier:
scratch this, found a better way

PR #220, opened 2021-09-22
Reduce the amount of logging when persisting nodes

hubertp:
Persisting nodes may produce a lot of logs which is not particularly
interesting when running frontends with INFO level.
This change is pretty subjective but in general that level of logging is
not necessary for regular users.


bbrehm (approved):
Very good.

I especially like that you left the  `"heap usage after GC: " + heapUsagePercent + "% -> will clear some references (if possible)";` message in -- that one is actually important: 

Most interactive users should kill the session and try again when encountering overflow; it is only good for the very situational case that you MUST run some analysis, you don't care if it takes multiple hours, you don't have enough memory to complete it and CANNOT acquire a box with more ram; and you have almost enough memory, such that the analysis can complete with overflow but cannot complete without overflow.

mpollmeier:
👍🏻

PR #226, opened 2021-11-04
Fix NPE in DiffTool

note:
Added `DiffToolTest`

`DiffToolTest.returnSomeDifferencesIfOneGraphHasMoreNodeProperties` was failing with the previous version of the code


mpollmeier (approved):
💯

PR #228, opened 2021-11-10
Make RepeatStep iterator reusable.

ml86:
Factored out the iterator class used in RepeatStep and made it
independent of Traversal class.
Furthermore i moved the work list creation into the iterator class to
make sure that no funny stuff can happen because of the shared work
list. The way it is currently used works but if the generated traversals
are not completely drained before the next traversal is created and
used, the work items get mixed up.

I piggy bagged changes to make RepeatBehaviour independent from Traversal and cleanup for the RepeatBehaviour Builder.


mpollmeier (approved):
good idea, good refactoring!

nit: there's a fair amount of pure formatting, it would have been nice to have that in a separate commit or even PR

PR #230, opened 2021-11-16
Executor service configuration.

Ferada:
Going forward, we'd like to set the executor from the outside, for, amongst other things, being able to correctly configure the logging context being used.

### DIFF
--- a/core/src/main/java/overflowdb/ReferenceManager.java
+++ b/core/src/main/java/overflowdb/ReferenceManager.java
@@ -26,7 +26,8 @@ public class ReferenceManager implements AutoCloseable, HeapUsageMonitor.HeapNot
 
   public final int releaseCount = 100000;
   private AtomicInteger totalReleaseCount = new AtomicInteger(0);
-  private final ExecutorService singleThreadExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory("overflowdb-reference-manager"));
+  private final ExecutorService executorService;
+  private final boolean shouldShutdown;
   private int clearingProcessCount = 0;
   private final Object backPressureSyncObject = new Object();
   private final OdbStorage storage;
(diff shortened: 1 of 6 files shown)


core/src/main/java/overflowdb/ReferenceManager.java
```diff
@@ -39,6 +37,15 @@
   public ReferenceManager(OdbStorage storage, NodesWriter nodesWriter) {
     this.storage = storage;
     this.nodesWriter = nodesWriter;
+    this.executorService = Executors.newSingleThreadExecutor(new NamedThreadFactory("overflowdb-reference-manager"));
+    this.shouldShutdown = false;
```
mpollmeier, line 41: why don't we want to shutdown the executorService if it's the default singlethreaded one? we used to, in the past...

mpollmeier (changes requested):

core/src/main/java/overflowdb/ReferenceManager.java
hubertp: If it is used outside of overflowdb that would be undesired

hubertp: but i guess it should be the other way around in the constructors, Olof?

Ferada: Yes, that should've been `true` of course :+1:

core/src/main/java/overflowdb/ReferenceManager.java
```diff
@@ -29,7 +26,8 @@
 
   public final int releaseCount = 100000;
   private AtomicInteger totalReleaseCount = new AtomicInteger(0);
-  private final ExecutorService singleThreadExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory("overflowdb-reference-manager"));
+  private final ExecutorService executorService;
+  private final boolean shouldShutdown;
```
mpollmeier, line 30: can we rename this to something more specific? e.g. `shutdownExecutorOnClose`?

mpollmeier (approved):
suggested a name change, apart from that 👍🏻

Ferada:
> suggested a name change, apart from that 👍🏻

Made the change and added a comment to the constructors to explain what the difference is.

PR #231, opened 2021-11-17
string filter steps: support multi line string (both generic and domain specific)

mpollmeier:


ml86 (commented):
What is wrong with people explicitly using `(?s)`?

ursachec (approved):

ursachec:
@ml86 the feature came as a request from my side, tbh mostly because i did not know about the `s` modifier. i still think it makes sense to have it as a default though because otherwise people would have a hard time writing their fire-and-forget regexes in CPGs with a lot of multi-lined-CODE-property CALL nodes (common case in CPGs coming out of the Ghidra decompiler via ghidra2cpg). Or, in cases in which you have a lot of single-lined-CODE CALL nodes visible on the screen, with a few which are further down the Joern shell when you wouldn't even expect multi-line matches to be necessary, leading to unexpected lost matches.

ml86:
It is one thing that we do not adhere to the standard regex anymore and now have to document: "Btw. we are always prefixing (?s) and you cannot turn that off".
But this also changes the semantics of already written regex. One example would be the greedy ".*" operator in existing queries which might unexpectedly match a lot more now.

PR #236, opened 2021-11-29
make 'close' synchronized

mpollmeier:
To ensure multiple callers don't trigger shutdown procedure twice.

### DIFF
--- a/core/src/main/java/overflowdb/Graph.java
+++ b/core/src/main/java/overflowdb/Graph.java
@@ -13,6 +13,7 @@ import overflowdb.util.PropertyHelper;
 
 import java.io.IOException;
 import java.util.*;
+import java.util.concurrent.*;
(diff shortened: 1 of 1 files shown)


ml86 (commented):
Not sure whether we really need the `volatile` but it wont hurt much.

ml86:
One more comment: I think we should not need the possibility to close from multiple threads but it wont hurt either.

mpollmeier:
`volatile` was a leftover from a previous, more complex approach - I just forgot to remove it - done now. 

The reason I'm proposing this is that we may have cases where multiple threads do exactly that, resulting in ConcurrentModificationException - e.g. https://github.com/ShiftLeftSecurity/product/issues/9364 and https://github.com/ShiftLeftSecurity/product/issues/9229

ml86 (approved):

PR #241, opened 2021-12-06
Adjust code parts which the Scala 3 compiler does not like.

ml86:


mpollmeier (approved):
as discussed in slack

PR #290, opened 2022-03-02
Improve BatchedUpdate.DiffGraphBuilder.addEdge API

bbrehm:
During porting of our giant collection of passes, I stumbled upon the problem that I naively assumed that the age (NewNode vs StoredNode) of src and dst nodes is always statically known. This is not true for many passes.

The old api functions are retained for binary compatibility. The new addEdge function takes Object and does runtime checks, because nobody needs union types, right?

Note that the scala API looked somewhat nicer, because there is a more sensible `AbstractNode` shared supertype of NewNode and StoredNode. Since these are not odb concepts, we cannot do this here.

### DIFF
--- a/core/src/main/java/overflowdb/BatchedUpdate.java
+++ b/core/src/main/java/overflowdb/BatchedUpdate.java
@@ -114,6 +114,15 @@ public class BatchedUpdate {
             return this;
         }
 
+        public DiffGraphBuilder addEdge(Object src, Object dst, String label, Object... properties) {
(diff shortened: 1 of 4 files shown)


mpollmeier:
I don't really like `Object` here and would much prefer to move `AbstractNode` (or a similar new marker trait) into ODB. Currently the odb-codegen generates it, but in reality it's a static class anyway, in the sense that it doesn't depend on the schema at all. 
What do you think?

bbrehm:
OK, will do.

mpollmeier (approved):

PR #292, opened 2022-03-10
Make Traversal and PathAwareTraversal covariant.

ml86:
The motivation for this are the current preparations for Scala 3 in down
stream projects which currently heavily rely on the implicit invocation of
`overflowdb.traversal.iterableToTraversal`. `iterableToTraversal` is in the
implicit resolution scope because it is defined in the `traversal`
package object. In Scala 3 functions defined there are no longer
considered to automatically be in scope and thus an explicit import of
the `iterableToTraversal` would be necessary.
To avoid this we can make the Traversal itself covariant like the
IterableOnce it wraps.

Sadly here comes a historic blunder into play which is the definition of
`toSet` on Traversal which does not override but rather coexists with
`toSet` defined on IterableOps from which Traversal derives. This was
possible because the IterableOps version is defined as
`def toSet[B >: A]: immutable.Set[B]` whereas the Traversal version was
defined as `def toSet: mutable.Set[A]`. With Traversal being covariant
in A we would need to change the Traversal version to
`def toSet[B >: A]: mutable.Set[B]` which is an unallowed overload in
only the method return type.
My solution to this is that we remove the Traversal version of `toSet`
and just provide the version coming from IterableOps which is a breaking
API change but one which only pops up if the changed return type
conflicts with the call site. At some point we anyway need to break this
since it is super unintuitive.

### DIFF
--- a/traversal/src/main/scala/overflowdb/traversal/PathAwareTraversal.scala
+++ b/traversal/src/main/scala/overflowdb/traversal/PathAwareTraversal.scala
@@ -3,7 +3,7 @@ package overflowdb.traversal
 import scala.annotation.tailrec
 import scala.collection.{IterableOnce, Iterator}
 
-class PathAwareTraversal[A](val elementsWithPath: IterableOnce[(A, Vector[Any])])
+class PathAwareTraversal[+A](val elementsWithPath: IterableOnce[(A, Vector[Any])])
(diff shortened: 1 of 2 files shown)


ml86:
Thats the usual price for a covariant type. E.g. `IterableOps` also has it all over the place.

bbrehm (approved):

PR #299, opened 2022-05-02
import/export neo4j csv

mpollmeier:
I didn't yet test this with an actual neo4j instance, and I'm not sure how this handles 
our 'special' `CONTAINED` edge, but this feels like a good point to wrap up a first PR. 
https://neo4j.com/docs/operations-manual/current/tools/neo4j-admin/neo4j-admin-import/


DavidBakerEffendi (approved):

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvExporter.scala
```diff
@@ -0,0 +1,92 @@
...
+  /** write edges of all labels */
+  private def exportEdges(graph: Graph, outputRootDirectory: Path): Seq[Path] = {
+    val edgeFilesContextByLabel = mutable.Map.empty[String, EdgeFilesContext]
+
+    graph.edges().forEachRemaining { edge =>
+      val label = edge.label
+      val context = edgeFilesContextByLabel.getOrElseUpdate(label, {
+        // first time we encounter an edge of this type - create the columnMapping and write the header file
+        val headerFile = outputRootDirectory.resolve(s"$label$HeaderFileSuffix.csv")  // to be written at the very end, with complete ColumnDefByName
+        val dataFile   = outputRootDirectory.resolve(s"$label.csv")
```
note, line 60: I noticed that we have the same naming convention for both edges and nodes. I don't know if labels are guaranteed to be unique across edges and nodes. If they're not then maybe we should e.g. `s"edge_$label.csv` or anything in this spirit?

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvExporter.scala
```diff
@@ -0,0 +1,92 @@
...
+    nodeFiles ++ edgeFiles
+  }
+
+  private def exportNodes(graph: Graph, label: String, outputRootDirectory: Path): Seq[Path] = {
+    val dataFile   = outputRootDirectory.resolve(s"$label.csv")
+    val headerFile = outputRootDirectory.resolve(s"$label$HeaderFileSuffix.csv")  // to be written at the very end, with complete ColumnDefByName
+    val columnDefinitions = new ColumnDefinitions(graph.nodes(label).next.propertyKeys.asScala)
+
+    Using(CSVWriter.open(dataFile.toFile, append = false)) { writer =>
+      Traversal(graph.nodes(label)).foreach { node =>
```
note, line 40: Any reason to prefer `Traversal(...).foreach` to:

```
graph.nodes(label).forEachRemaining { node =>
        val specialColumns = Seq(node.id.toString, node.label)
        val propertyValueColumns = columnDefinitions.propertyValues(node.propertyOption(_).toScala)
        writer.writeRow(specialColumns ++ propertyValueColumns)
      }
```

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
```diff
@@ -0,0 +1,244 @@
...
+        throw new AssertionError(s"unable to determine file type - neither ${ColumnType.Label} (for nodes) nor ${ColumnType.Type} (for relationships) found")
+    }
+  }
+
+  private def parseNodeRowData(columnsRaw: Seq[String], lineNo: Int, columnDefs: Map[Int, ColumnDef]): ParsedNodeRowData = {
+    var id: Integer = null
+    var label: String = null
+    val properties = Seq.newBuilder[ParsedProperty]
+    columnsRaw.zipWithIndex.foreach { case (entry, idx) =>
+      assert(columnDefs.contains(idx), s"column with index=$idx not found in column definitions derived from headerFile")
```
note, line 146: I would consider:

```
      columnDefs.get(idx) match {
        case Some(ColumnDef(_, ColumnType.Id, _)) =>
          id = entry.toInt
        case Some(ColumnDef(_, ColumnType.Label, _)) =>
          label = entry
        case Some(ColumnDef(Some(name), valueType, false)) =>
          parseProperty(entry, name, valueType).foreach(properties.addOne)
        case Some(ColumnDef(Some(name), valueType, true)) =>
          parseArrayProperty(entry, name, valueType).foreach(properties.addOne)
        case Some(other) =>
          throw new MatchError(s"unhandled case $other in $lineNo: ${columnsRaw.mkString(",")}")
        case None =>
          throw new AssertionError(s"column with index=$idx not found in column definitions derived from headerFile")
      }
```

It's bit more idiomatic for my taste. Also, it uses just one Map lookup instead of two (`Map.contains` and `Map.apply` both need to look up for the key)

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
```diff
@@ -0,0 +1,244 @@
...
+      columnDefs(idx) match {
+        case ColumnDef(_, ColumnType.Id, _) =>
+          id = entry.toInt
```
note, line 149: How about:

```
Try(entry.toInt).fold(err => throw new AssertionError(s"ColumnType.Id is not an Int ($lineNo): $entry: $err"), identity(_))
```
in every place we use `toInt`? Mostly to provide more context, like line number, in the error messages

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
```diff
@@ -0,0 +1,244 @@
...
+        case ColumnDef(_, ColumnType.Label, _) =>
+          label = entry
+        case ColumnDef(Some(name), valueType, false) =>
+          parseProperty(entry, name, valueType).foreach(properties.addOne)
+        case ColumnDef(Some(name), valueType, true) =>
+          parseArrayProperty(entry, name, valueType).foreach(properties.addOne)
+        case other =>
+          throw new MatchError(s"unhandled case $other")
```
note, line 157: scala```
throw new MatchError(s"unhandled case $other in $lineNo")
```

I would consider adding `filename` to `parseNodeRowData` just to provide more context in error messages

File: formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala [hunk 6]
```diff
@@ -0,0 +1,244 @@
...
+    assert(endId != null, s"no END_ID column found in line $lineNo")
+    assert(label != null, s"no LABEL column found in line $lineNo")
+
+    val ret = ParsedEdgeRowData(startId, endId, label, properties.result())
+    logger.debug("parsed line {}: {}", lineNo, ret)
+    ret
+  }
+
+  private def parseProperty(rawValue: String, name: String, valueType: ColumnType.Value): Option[ParsedProperty] = {
+    if (rawValue != "" || valueType == ColumnType.String)
```
note, line 200: Is `valueType == ColumnType.String` some temporary check? Asking because here we're very strict about `valueType`, basically filtering in only String, while `parsePropertyValue` actually handles much more types

note (approved):

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
```diff
@@ -0,0 +1,244 @@
...
+      Some(ParsedProperty(name, parsedValues))
+    } else {
+      None
+    }
+  }
+
+  private def parsePropertyValue(rawString: String, valueType: ColumnType.Value): Any = {
+    valueType match {
+      case ColumnType.Int => rawString.toInt
+      case ColumnType.Long => rawString.toLong
```
note, line 219: In all those `toX` the same as https://github.com/ShiftLeftSecurity/overflowdb/pull/299/files#r863641990 applies

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvExporter.scala
mpollmeier: good point, done

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvExporter.scala
mpollmeier: not really, and yes you're right, using the simplest tool for the job would be the right thing to do. done

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
mpollmeier: Hmm, I started with what you're proposing here, but then changed it around to the current version, which I found more readable. There's already a lot of clutter in decomposing `ColumnDef`, if you add `Option` to the mix it starts getting a bit hard to digest. Matter of taste I guess. 

tl;dr; I see your point but will give this one a miss :)

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
mpollmeier: yes I'm lazy, and you're right... done :)

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
mpollmeier: done

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
mpollmeier: This is a bit confusing - I added a comment in b9546ec5f84c47ffe26e7870fc7b6b841c67ee67, hopefully that makes more sense?
```
    // an empty string is only a valid value if the valueType is String - otherwise it simply indicates an unset property
```

mpollmeier:
@note i made quite a few changes re parsing, based on your PR suggestions... would you mind having another look?

formats/src/main/scala/overflowdb/formats/neo4jcsv/Neo4jCsvImporter.scala
note: Yes, it makes sense. Actually my initial conclusion came from me misreading the condition, I've taken `||` as `&&`.

note:
@mpollmeier Thanks for applying comments. It looks good to me

PR #300, opened 2022-05-04
odb importer/exporter as (abstract) main apps

mpollmeier:
This allows us to create simple CLI apps to import/export from/to different standard graph formats like neo4jcsv. 
We cannot make them entirely generic unfortunately, because at runtime they depend on the (usually generated) domain specific code.


formats/src/main/scala/overflowdb/formats/ExporterMainBase.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.jdk.CollectionConverters.SeqHasAsJava
+import scala.util.Using
+
+/**
+ * Main class to export a given OverflowDB graph to various export formats.
+ * Because ODB relies on domain specific implementations, this is an abstract class - to use it, we need the
+ * NodeFactories and EdgeFactories of the specific domain implementation, e.g. from the classes generated
+ * by https://github.com/ShiftLeftSecurity/overflowdb-codegen
+ */
+abstract class ExporterMainBase extends App {
```
note, line 20: I know it is subjective but to me abstract class + `App` looks very involving for a task at hand.

How about:

```scala
object ExporterMainBase {
  def export(def export(nodeFactories: Seq[NodeFactory[_]], edgeFactories: Seq[EdgeFactory[_]])(args: Array[String]): Unit = {
    val logger = LoggerFactory.getLogger(getClass)
    ...
}
```

Then in `Neo4jCsvTests` you can just:

```scala
val testExport = ExporterMainBase.export(Seq(TestNode.factory), Seq(TestEdge.factory))
testExport(Array("--format=neo4jcsv", s"--out=${exportPath.pathAsString}", graphPath.pathAsString))
testExport(somethingElse)
```

I think it's simpler as it uses less concepts, mostly just functions calls, instead of inheritance and `App`, which bring `DelayedInit` and the order of execution is getting complicated

note (approved):

formats/src/main/scala/overflowdb/formats/ExporterMainBase.scala
mpollmeier: I wanted to make the usage site as simple as possible, i.e. when we generate/write/package these main apps for specific domains, e.g. cpg. 

But yeah, I think you're right, `App` brings in a lot of complexity, and it's probably less confusing to be explicit about it's usage anyway, even if that involves a few more lines when using it.

mpollmeier: done

PR #301, opened 2022-05-13
write cypher files with import column mappings

mpollmeier:
Neo4j has several exchange formats (a few csvs, graphml, ...), but none of them seem to work... 
When I followed the (well-written) docs I couldn't find a single format that even passed a simple
round-trip test, i.e. export and import, even using their test data. 

Instead, our exporter now creates a separate file for each node|edge type: the cypher query to read the `_data.csv`
as a csv and map it's columns to the correct properties, converting them to the correct value types.


formats/src/main/scala/overflowdb/formats/ImporterMain.scala
```diff
@@ -34,7 +37,10 @@ object ImporterMain extends App {
             case Format.GraphMl => GraphMLImport
           }
           val odbConfig = overflowdb.Config.withoutOverflow.withStorageLocation(outputFile)
-          Using.resource(Graph.open(odbConfig, nodeFactories.asJava, edgeFactories.asJava)) { graph =>
+          val convertPropertyForPersistenceJava = new function.Function[Object, Object] {
+            override def apply(o: Object) = convertPropertyForPersistence(o).asInstanceOf[Object]
+          }
+          Using.resource(Graph.open(odbConfig, nodeFactories.asJava, edgeFactories.asJava, convertPropertyForPersistenceJava)) { graph =>
```
note, line 43: Nit:

```
Using.resource(Graph.open(odbConfig, nodeFactories.asJava, edgeFactories.asJava, (o: Object) => convertPropertyForPersistence(o).asInstanceOf[Object]))```

also seems to work, i.e. no need to use `java.function.Function` explicitly

mpollmeier: true! even better IMO: `convertPropertyForPersistence(_).asInstanceOf[Object]`

PR #302, opened 2022-05-19
graphml as import/export format

mpollmeier:
Note: GraphML doesn't natively support list property types, so we fake it by encoding it as a `;` delimited string.
If you import this into a different database, you'll need to parse that separately.
In comparison, Tinkerpop just bails out if you try to export a list property to graphml.

### DIFF
--- a/formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
+++ b/formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
@@ -7,6 +7,7 @@ import java.lang.System.lineSeparator
 import java.nio.file.{Files, Path}
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.{IterableHasAsScala, IteratorHasAsScala, MapHasAsScala}
+import scala.xml.{PrettyPrinter, XML}
 
 /**
  * Exports OverflowDB Graph to GraphML
@@ -67,8 +68,8 @@ object GraphMLExporter extends Exporter {
        |    </graph>
        |</graphml>
        |""".stripMargin.strip
-
     Files.writeString(outFile, xml)
+    xmlFormatInPlace(outFile)
 
     ExportResult(
       nodeCount = nodeEntries.size,
@@ -149,4 +150,12 @@ object GraphMLExporter extends Exporter {
       case _ => value.toString
     }
   }
+
+  private def xmlFormatInPlace(xmlFile: Path): Unit = {
+    val xml = XML.loadFile(xmlFile.toFile)
+    val prettyPrinter = new PrettyPrinter(120, 2)
+    val formatted = prettyPrinter.format(xml)
+    Files.writeString(xmlFile, formatted)
+  }
+
 }
--- a/formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
+++ b/formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
@@ -5,7 +5,7 @@ import overflowdb.{Element, Graph}
 
 import java.nio.file.{Files, Path}
 import scala.collection.mutable
-import scala.jdk.CollectionConverters.{IteratorHasAsScala, MapHasAsScala}
+import scala.jdk.CollectionConverters.{IterableHasAsScala, IteratorHasAsScala, MapHasAsScala}
 
 /**
  * Exports OverflowDB Graph to GraphML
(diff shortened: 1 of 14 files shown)


formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
```diff
@@ -0,0 +1,151 @@
...
+      val tpe0 =
+        if (isList) s"$tpe[]"
+        else tpe
+      s"""<key id="$key" for="node" attr.name="$name" attr.type="$tpe0"></key>"""
+    }.mkString("\n")
+    val edgePropertyKeyEntries = edgePropertyContextById.map { case (key, PropertyContext(name, tpe, isList)) =>
+      val tpe0 =
+        if (isList) s"$tpe[]"
+        else tpe
+      s"""<key id="$key" for="edge" attr.name="$name" attr.type="$tpe"></key>"""
```
note, line 53: Should be `tpe0`

formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
```diff
@@ -0,0 +1,151 @@
...
+    }.mkString("\n")
```
note, line 54: `System.lineSeparator` would be better in case someone runs it on Windows

formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
```diff
@@ -0,0 +1,151 @@
...
+
+    val edgeEntries = graph.edges().asScala.map { edge =>
+      s"""<edge source="${edge.outNode.id}" target="${edge.inNode.id}">
+         |    <data key="$KeyForEdgeLabel">${edge.label}</data>
+         |    ${dataEntries("edge", edge, edgePropertyContextById)}
+         |</edge>
+         |""".stripMargin
+    }.toSeq
+
+    val nodePropertyKeyEntries = nodePropertyContextById.map { case (key, PropertyContext(name, tpe, isList)) =>
```
note, line 43: After understanding what it does, how about:

```scala
def propertyKeyXml(forAttr: String, propsMap: mutable.Map[String, PropertyContext]): String = {
      propsMap.map { case (key, PropertyContext(name, tpe, isList)) =>
        val tpe0 =
          if (isList) s"$tpe[]"
          else tpe
        s"""<key id="$key" for="$forAttr" attr.name="$name" attr.type="$tpe0"></key>"""
      }.mkString(System.lineSeparator())
    }

val nodePropertyKeyEntries = propertyKeyXml("node", nodePropertyContextById)
val edgePropertyKeyEntries = propertyKeyXml("edge", nodePropertyContextById)
```

formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
```diff
@@ -0,0 +1,151 @@
...
+       |    <graph id="G" edgedefault="directed">
+       |      ${nodeEntries.mkString("\n")}
+       |      ${edgeEntries.mkString("\n")}
+       |    </graph>
+       |</graphml>
+       |""".stripMargin.strip
+
+    Files.writeString(outFile, xml)
+
+    ExportResult(
```
note, line 74: I would use named parameters for all params

note (approved):
https://github.com/ShiftLeftSecurity/overflowdb/pull/302#discussion_r877099327 is the only major comment

formats/src/main/scala/overflowdb/formats/graphml/GraphMLImporter.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  private def parsePropertyEntries(forElementType: String, keyEntries: NodeSeq): Map[String, PropertyContext] = {
+    keyEntries
+      .filter(_ \@ "for" == forElementType)
+      .map { node =>
+        val id = node \@ "id"
+        val name = node \@ "attr.name"
+        val graphmlType = node \@ "attr.type"
+
+        // warning: this is derivating from the graphml spec - we do want to support list properties...
```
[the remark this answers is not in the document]
mpollmeier: my english not soo gut 😊

formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
mpollmeier: done

formats/src/main/scala/overflowdb/formats/graphml/GraphMLExporter.scala
mpollmeier: fyi copy pasta: the second context should be `edgePropertyContextById`

mpollmeier:
fyi I'll keep this open until testing this with different graph dbs/tools has completed.

PR #304, opened 2022-05-24
export to graphviz dot format

mpollmeier:
Leaving out the importer for now - mostly because dot is untyped... we could generate 
'importFromString' methods for each property in the odb-codegen in the future...

Note: GraphML doesn't natively support list property types, so we fake it by encoding it as a `;` delimited string.

Random links:
* https://en.wikipedia.org/wiki/DOT_(graph_description_language)
* https://www.graphviz.org/doc/info/lang.html
* http://magjac.com/graphviz-visual-editor/

### DIFF
--- a/formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
+++ b/formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
@@ -43,8 +43,6 @@ object DotExporter extends Exporter {
       }
 
       writeLine("}")
-      writer.flush()
-      writer.close()
     }
 
     ExportResult(
@@ -87,19 +85,4 @@ object DotExporter extends Exporter {
     }
     outputRootDirectory.resolve("export.dot")
   }
-
-  private def encodeListValue(value: AnyRef): String = {
-    value match {
-      case value: Iterable[_] =>
-        value.mkString(";")
-      case value: IterableOnce[_] =>
-        value.iterator.mkString(";")
-      case value: java.lang.Iterable[_] =>
-        value.asScala.mkString(";")
-      case value: Array[_] =>
-        value.mkString(";")
-      case _ => value.toString
-    }
-  }
-
 }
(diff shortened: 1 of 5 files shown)


formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
```diff
@@ -0,0 +1,105 @@
...
+  private def resolveOutputFile(outputRootDirectory: Path): Path = {
+    if (Files.exists(outputRootDirectory)) {
+      assert(Files.isDirectory(outputRootDirectory), s"given output directory `$outputRootDirectory` must be a directory, but isn't...")
+    } else {
+      Files.createDirectories(outputRootDirectory)
+    }
+    outputRootDirectory.resolve("export.dot")
+  }
+
+  private def encodeListValue(value: AnyRef): String = {
```
note, line 91: IntelliJ marks this method as unused

formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      graph.edges().forEachRemaining { edge =>
+        edgeCount += 1
+        writeLine(edge2Dot(edge))
+      }
+
+      writeLine("}")
+      writer.flush()
+      writer.close()
```
note, line 47: Technically you don't have to `close` since it's withing `Using.resource` block

formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
mpollmeier: true, i solved this differently. deleted now

formats/src/main/scala/overflowdb/formats/dot/DotExporter.scala
mpollmeier: very true, deleted, including the `flush` which happens on close anyway

note (approved):

PR #307, opened 2022-05-26
TraversalSource:id should default node type to `Node`

mpollmeier:
rather than `Nothing` as before. Otherwise, this invalid traversal
would compile: `graph.id(0).outV`. Now, we get a compiler error.

Re https://github.com/joernio/joern/issues/1367

### DIFF
--- a/traversal/src/main/scala/overflowdb/traversal/DefaultsToNode.scala
+++ b/traversal/src/main/scala/overflowdb/traversal/DefaultsToNode.scala
@@ -13,7 +13,7 @@ object DefaultsToNode {
   implicit def overrideDefault[A]: DefaultsToNode[A] =
     new DefaultsToNode[A]
 
-  implicit def default: DefaultsToNode[Node] =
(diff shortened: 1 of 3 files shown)


note (approved):

traversal/src/main/scala/overflowdb/traversal/DefaultsToNode.scala
```diff
@@ -0,0 +1,20 @@
...
+ * used e.g. for `NodeTypeStarters:id`
+ * */
+sealed class DefaultsToNode[A]
+
+object DefaultsToNode {
+
+  implicit def overrideDefault[A]: DefaultsToNode[A] =
+    new DefaultsToNode[A]
+
+  implicit def default: DefaultsToNode[Node] =
```
note, line 16: Could be `val`?

mpollmeier: true. done

PR #309, opened 2022-06-03
Allow more simple constructions to use the index

bbrehm:
add new Traversal subclass for initial traversals over the graph. Idea is that cpg.method.fullNameExact should use the index instead iterating over everything. Hence, cpg.method needs to be the new type of traversal, and the implicit .fullNameExact needs to use reflection to check whether it can potentially use the index. However, we can only do this if the iterator itself is virgin, e.g. val trav = cpg.method; trav.next; trav.fullNameExact(...) cannot use the index.

The way this will look like in the implicit is 
```
  def fullNameExact(value: String): Traversal[NodeType] = {
val fast = traversal match{
case init: InitialTraversal[A] => init.getByIndex("FULL_NAME", fn)
case _ => null
}
if(fast != null) fast  else  traversal.filter { node => node.fullName == value }
}
```

The main reason is that lots of users -- especially our internal users who write framework code -- can't be convinced to consistently use the TraversalSource classes in their code. It always looks like `cpg.method.fullName("foo")`.

Follow-ups to make this useful:
1. NodeStarters must generate the new InitialTraversal
2. overflowdb-codegen must emit the code to check this.

### DIFF
--- a/traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
+++ b/traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
@@ -1,0 +1,38 @@
+package overflowdb.traversal
+
+import overflowdb.Graph
+
+class InitialTraversal[+A <: overflowdb.Node] private (graph: Graph,
+                                                       label: String,
+                                                       iter: ArrayListIter[A])
+    extends Traversal[A](iter) {
+
+  def getByIndex(key: String, value: Any): Traversal[A] = {
+    if (iter.idx == 0 && graph.indexManager.isIndexed(key)) {
+      val nodes = graph.indexManager.lookup(key, value)
+      Traversal.from(nodes.iterator()).asInstanceOf[Traversal[A]]
+    } else {
+      null
+    }
+  }
+}
+
+object InitialTraversal {
+  def from[A](graph: Graph, label: String): InitialTraversal[A] = {
+    new InitialTraversal(
+      graph,
+      label,
+      new ArrayListIter[A](graph.nodes.nodesByLabel(label).asInstanceOf[java.util.ArrayList[A]]))
+  }
+}
+private[overflowdb] class ArrayListIter[+T](arr: java.util.ArrayList[T]) extends Iterator[T] {
+  private[overflowdb] var idx = 0
+
+  override def hasNext: Boolean = idx < arr.size()
+
+  override def next(): T = {
+    if (!hasNext) throw new NoSuchElementException()
+    idx = idx + 1
+    arr.get(idx - 1)
+  }
+}
--- a/core/src/main/java/overflowdb/Graph.java
+++ b/core/src/main/java/overflowdb/Graph.java
@@ -21,7 +21,7 @@ public final class Graph implements AutoCloseable {
   private static final Logger logger = LoggerFactory.getLogger(Graph.class);
 
   protected final AtomicLong currentId = new AtomicLong(-1L);
-  protected final NodesList nodes = new NodesList();
+  final NodesList nodes = new NodesList();
   public final IndexManager indexManager = new IndexManager(this);
   private final Config config;
   private boolean closed = false;


mpollmeier:
related: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1453 - sorry I forgot about that one

traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import overflowdb.Graph
+
+class InitialTraversal[+A <: overflowdb.Node] private (graph: Graph,
+                                                       label: String,
+                                                       iter: ArrayListIter[A])
+    extends Traversal[A](iter) {
+
+  def getByIndex(key: String, value: Any): Traversal[A] = {
+    if (iter.idx == 0 && graph.indexManager.isIndexed(key)) {
```
mpollmeier, line 11:
```suggestion
    // we can only do this if the iterator itself is virgin, e.g. `val trav = cpg.method; trav.next; trav.fullNameExact(...)` cannot use the index
    if (iter.idx == 0 && graph.indexManager.isIndexed(key)) {
```

traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
```diff
@@ -0,0 +1,38 @@
...
+      val nodes = graph.indexManager.lookup(key, value)
+      Traversal.from(nodes.iterator()).asInstanceOf[Traversal[A]]
+    } else {
+      null
```
mpollmeier, line 15: can we plesae avoi `null` here? I don't want to check for `null` all over the place like in old style java. `None` might be the best option here (pun intended)

bbrehm:
Hah, yeah. This is exactly the same approach I suggested back then.

traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
```diff
@@ -0,0 +1,40 @@
...
+
+class InitialTraversal[+A <: overflowdb.Node] private (graph: Graph,
+                                                       label: String,
+                                                       iter: ArrayListIter[A])
+    extends Traversal[A](iter) {
+
+  def getByIndex(key: String, value: Any): Traversal[A] = {
+    if (iter.idx == 0 && graph.indexManager.isIndexed(key)) {
+      val nodes = graph.indexManager.lookup(key, value)
+      Traversal.from(nodes.iterator()).asInstanceOf[Traversal[A]]
```
mpollmeier, line 13:
```suggestion
      Traversal.from(nodes.iterator()).cast[A]
```

traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
bbrehm: This interface is not user-facing, and None doesn't fit the desired type.

We already have a user-facing interface for this functionality (based on TraversalSource).

mpollmeier: > This interface is not user-facing

I am also a user


> None doesn't fit the desired type.

I don't follow. This method either returns a traversal, or not. Ever since java 8 and scala 2.x we can thankfully encapsulate this in the return type, in this case e.g. `Option[Traversal[A]]`. If we pass around `null` we have to check it all over the place, and it defers from common practices, not only for scala but also for java these days.

bbrehm: This is not intended to be a widely used public API; this function has exactly one intended use-site (in the generated code) that immediately checks the result for nullity, so there is no issue about forgetting a check or "passing around `null`".

If this specific case is important to you, then I can change it?

mpollmeier: thank you, yes please change it.

mpollmeier (approved):
i'll pre-approve to not hold you up any longer

PR #316, opened 2022-07-14
StringPropertyFilter.exactMultiple helper for varargs queries

maltek:
We discussed the bad performance of the varargs queries, so here's my attempt to fix that. But I haven't actually witnessed performance problems so this is mostly based on assumptions.

- use index if possible
- use a mutable HashSet instead of immutable
- Lazily create the HashSet only on the second iteration. Definitely a win when there's zero iterations. Since single-iteration is relatively common with `.where` and friends it seemed worth it to do delay set creation a bit more and take the potential hit of a single sequential scan.
- I thought about not using sets at all when we only have very few values to search for but some small bit of microbenchmarking convinced me that's a bad idea.

### DIFF
--- a/traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
+++ b/traversal/src/main/scala/overflowdb/traversal/InitialTraversal.scala
@@ -7,9 +7,11 @@ class InitialTraversal[+A <: overflowdb.Node] private (graph: Graph,
                                                        iter: ArrayListIter[A])
     extends Traversal[A](iter) {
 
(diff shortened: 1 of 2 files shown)


mpollmeier:
> Markus told me not to use the varargs methods because they are slow and bad

I would have been surprised if this has any significant impact on the overall sptests performance, but it's good to have anyway :)

> Should I add a comment saying that this is not worth it?

Yes please, I think adding that info from the last bullet point would be good to have as a comment. Other than that, I'm 👍🏻 for this.

mpollmeier (approved):

ml86:
Than lets merge this. After all it is in a very encapsulated place and it wont hurt.

PR #332, opened 2022-10-26
exporters: consolidate logic for output files

mpollmeier:
for all 'single-outputfile-exporters':
  if the given outputFile is a directory: export into a new file in that directory
  otherwise: use the given outputFile as is, and create all parent directories (if not there already)

also: make each exporter define it's default file extension
... so that we can use exporters generically from the outside, while using
their specific correct file extensions for the resulting files

### DIFF
--- a/formats/src/test/scala/overflowdb/formats/graphson/GraphSONTests.scala
+++ b/formats/src/test/scala/overflowdb/formats/graphson/GraphSONTests.scala
@@ -66,6 +66,7 @@ class GraphSONTests extends AnyWordSpec {
         exportResult.nodeCount shouldBe 3
         exportResult.edgeCount shouldBe 2
         val Seq(graphMLFile) = exportResult.files
+        println(graphMLFile)
 
         // import graphml into new graph, use difftool for round trip of conversion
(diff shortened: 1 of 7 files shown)


mpollmeier:
cc @DavidBakerEffendi

formats/src/test/scala/overflowdb/formats/graphson/GraphSONTests.scala
```diff
@@ -66,6 +66,7 @@ class GraphSONTests extends AnyWordSpec {
         exportResult.nodeCount shouldBe 3
         exportResult.edgeCount shouldBe 2
         val Seq(graphMLFile) = exportResult.files
+        println(graphMLFile)
```
DavidBakerEffendi, line 69: @mpollmeier an accidentally forgotten print?

fabsx00 (approved):

formats/src/test/scala/overflowdb/formats/graphson/GraphSONTests.scala
mpollmeier: indeedly, thanks, dropped now

PR #355, opened 2023-02-20
prevent duplication of nodeRef on deserialization

bbrehm:
As we saw from https://github.com/ShiftLeftSecurity/overflowdbv2/pull/11 , we have a problem when deserializing nodes: There are too many and too heavy (https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/205) NodeRefs.

This addresses the "too many" part. The issue is the following: When we deserialize a node, then we ask NodeFactory to please give us a new NodeDb object with the data. But the API helpfully also allocates a NodeRef object that is kept alive by the NodeDb!

So the pointer structure when deserializing looks like this:
```
NodeRef1 -(`private Node node`)-> NodeDb1 -(`private NodeRef ref`)-> NodeRef2 -(`private Node node`)-> NodeDb1
```
This is of course nonsense. We fix this by passing pre-existing NodeRef pointers into the `createNode` function and only allocate a new `NodeRef` when that is `null`.

This should save one NodeRef per node, which weighs 32 bytes (with https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/205; before it was 40 bytes) .

PS. can you add some more eyes that I got the referenceManager queue / registerRef right? We never want to add a nodeRef twice, nor do we want to forget it.

### DIFF
--- a/core/src/main/java/overflowdb/NodeFactory.java
+++ b/core/src/main/java/overflowdb/NodeFactory.java
@@ -7,8 +7,12 @@ public abstract class NodeFactory<V extends NodeDb> {
 
   public abstract NodeRef<V> createNodeRef(Graph graph, long id);
 
-  public V createNode(Graph graph, long id) {
-    final NodeRef<V> ref = createNodeRef(graph, id);
+  public  V createNode(Graph graph, long id) {
+  return createNode(graph, id, null);
+  }
+
+  public V createNode(Graph graph, long id, NodeRef<V> ref) {
+    if(ref == null) ref = createNodeRef(graph, id);
     final V node = createNode(ref);
     node.markAsDirty(); //freshly created, i.e. not yet serialized
     ref.setNode(node);
--- a/core/src/main/java/overflowdb/NodeFactory.java
+++ b/core/src/main/java/overflowdb/NodeFactory.java
@@ -8,7 +8,7 @@ public abstract class NodeFactory<V extends NodeDb> {
   public abstract NodeRef<V> createNodeRef(Graph graph, long id);
 
   public  V createNode(Graph graph, long id) {
-  return createNode(graph, id, null);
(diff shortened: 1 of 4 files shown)


core/src/main/java/overflowdb/NodeFactory.java
```diff
@@ -7,8 +7,12 @@
 
   public abstract NodeRef<V> createNodeRef(Graph graph, long id);
 
-  public V createNode(Graph graph, long id) {
-    final NodeRef<V> ref = createNodeRef(graph, id);
+  public  V createNode(Graph graph, long id) {
+  return createNode(graph, id, null);
```
mpollmeier, line 11:
```suggestion
    return createNode(graph, id, null);
```

core/src/main/java/overflowdb/NodeRef.java
```diff
@@ -99,9 +99,9 @@ private final synchronized N getSynchronized() throws IOException {
     if (ref != null) {
       return ref;
     } else {
-      final N node = readFromDisk(id);
+      final N node = readFromDisk();
       if (node == null) throw new IllegalStateException("unable to read node from disk; id=" + id);
-      this.node = node;
+      assert(this.node == node);
```
mpollmeier, line 104: Java's `assert`s are not being checked at runtime by default 😢 
I tend to `throw new AssertionError` in this case, but I guess there's other options. 

Also: the error text in this case wouldn't be very helpful.

mpollmeier:
Hmm, this doesn't seem to fix the underlying issue yet. I tested with https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/205 and still got twice as many Refs compared to the Db nodes...

mpollmeier (approved):
please fix the assert issue before merging

PR #360, opened 2023-04-13
add unit test to document (counter-intutive?) collectAll path-tracking behavior

bbrehm:
This documents the imo counter-intuitive path-tracking behavior of collectAll: The step is part of the path.

collectAll is used in our steps like `isMethod`. I think these should behave like filters, i.e. don't add to the path.

What do you think?

If we decide to change this, I would do that in a follow-up and first merge the unit-test that documents current behavior.

### DIFF
--- a/traversal-tests/src/test/scala/overflowdb/traversal/PathTraversalTests.scala
+++ b/traversal-tests/src/test/scala/overflowdb/traversal/PathTraversalTests.scala
@@ -57,6 +57,11 @@ class PathTraversalTests extends AnyWordSpec with ExampleGraphSetup {
           Seq(center, r1, "R1"))
       }
 
+      "collectAll: Do we really want to include intermediate reulsts?" in {
+        centerTrav.enablePathTracking.collectAll[Thing].path.toList shouldBe List(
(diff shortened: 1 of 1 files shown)


ml86:
Path tracking should record expansions from node to node not sets of nodes after each step.

bbrehm:
It does record expansions from node to node. The question is whether `trav.isMethod` is formally viewed as an expansion (i.e. the resulting path will have two subsequent entries for the method; this is current behavior) or whether it rather is formally viewed as a filter, i.e. equivalent to `trav.filter{_.isInstanceOf[Method]}` (i.e. the resulting path will have only a single entry for the method).

mpollmeier (commented):
I agree, `collectAll` is a filter step that shouldn't be tracked as a separate path entry. I.e `cpg.all.enablePathTracking.isMethod.isMethod.isMethod.paths` should contain paths of length 1.

the only reason i didn't hit 'approve' yet is the wording of the test :)

mpollmeier (approved):

ml86:
Should count as filter.

PR #361, opened 2023-04-14
collectAll path tracking

bbrehm:
To be merged after https://github.com/ShiftLeftSecurity/overflowdb/pull/360 (I want an unsquashed commit in our history that documents/tests current undesired behavior).

The code may superficially look more complicated than the old code. This is not the case; in fact the code is now much simpler: I unsugared the implicit classtag and its `unapply` method from the match statement.

The old documentation was somewhat optimistic, referring to "collect ... of type". I clarified that -- collectAll refers to the erased java (runtime) class, not the scala (compile-time) type.


ml86 (approved):
Not very scalaish, but i also like it better.

mpollmeier (approved):
👍🏻

PR #364, opened 2023-04-18
format everything

bbrehm:
I don't particularly care about the formatting. However, I absolutely hate when I do a small edit in a file and intellij reformats the entire file, and I then have to do extensive editing sessions to get a clean diff. Hence I'd like to get the codebase to a state where it is a fixed point of `sbt scalafmt`, and where I don't have to manually format my PRs (so not the trivial fixed point of an invalid `.scalafmt.conf`).

If you have ideas how to configure scalafmt such that we don't have a giant diff (other dialect? other version?) then I'd be happy for proposals.


mpollmeier (approved):
I just sneakily pushed two changes onto this PR: 
1) upgrade to latest scalafmt (just because)
2) enforce scalafmtCheck on PR (I know it's a pain, but this way we can avoid another large PR like this in the future, no silver bullets here)...

I.e. if you agree with the above, just merge this.

PR #365, opened 2023-04-18
wip: remove traversal

bbrehm:
This gets rid of the Traversal class. WIP because we need to check how bad this is to bring into everything.


traversal-tests/src/test/scala/overflowdb/traversal/PathTraversalTests.scala
```diff
@@ -200,7 +202,7 @@ class PathTraversalTests extends AnyWordSpec with ExampleGraphSetup {
           Seq(center, r1, r2)
         )
         // we can hide internal steps from path-tracking
-        centerTrav.enablePathTracking.union(t => Traversal.from(t.out.out).iterator).path.toSetMutable shouldBe Set(
+        centerTrav.enablePathTracking.union(t => (t.out.out.l).iterator).path.toSetMutable shouldBe Set(
```
mpollmeier, line 205:
```suggestion
        centerTrav.enablePathTracking.union(_.out.out.iterator).path.toSetMutable shouldBe Set(
```
n.b. i didn't actually run/test this

mpollmeier (approved):

bbrehm:
Superseded by https://github.com/ShiftLeftSecurity/overflowdb/pull/367

traversal-tests/src/test/scala/overflowdb/traversal/PathTraversalTests.scala
bbrehm: no, the point of this test is that we hide the internal state by collecting the entire iterator and creating a new one. The variant you wrote will not hide the intermediate steps because `.out` uses `.flatMap` which is overridden in PathTrackingTraversal.

PR #366, opened 2023-04-20
github actions: upgrade to ubuntu 22.04

mpollmeier:
ubuntu 18.04 got deprecated
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,7 +2,7 @@ name: PR
 on: pull_request
 jobs:
   pr:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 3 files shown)


mpollmeier:
scalafmt ftw :)

image

I'll better disable it in that file...

bbrehm:
Yeah, it is far too aggressive. I'd much prefer a setting that is more non-invasive. For example, don't change line-breaks except for the last line, only change whitespace within lines.

bbrehm (approved):

PR #374, opened 2023-06-01
remove traversal compat

bbrehm:
This is upstream of https://github.com/joernio/joern/pull/2784


mpollmeier (approved):
the diff would have been smaller without the rename iterator -> iter :)

bbrehm:
>the diff would have been smaller without the rename iterator -> iter :)

That's correct. But there is the `iterator` function in the `Iterator` and `IterableOnce` interfaces. There is no actual name clash (since we're not implementing/extending either interface, we're just AnyVal-wrapping it), but it is close enough that `iterator` is a terrible name for the wrapped member: Better to anticipate all possible confusions and avoid them.

mpollmeier:
true that!

PR #414, opened 2024-02-05
diffgraph: ensure we don't crash if a node get's deleted twice

mpollmeier:


bbrehm:
:+1: 

Do we want to emit a warning?

I think this PR was triggered by a recent codescience crash, where the double-deletion was indicative of a logical / pipeline-config bug (and that only led to much later NPE crashes far removed from the source of the problem).

OTOH, there is no a-priori reason to forbid double-deletion in batched updates, so maybe a warning would be overly restrictive.

bbrehm (approved):

maltek:
Michael already merged, but since I've typed this already my 2cts:


> Do we want to emit a warning?

IMO either this is a supported use-case, or not. A warning would be a weird in-between. The way that (non-buggy) passes deal with the current restriction is by doing the same thing as here: keep a hash-set in order to not queue the same thing for deletion twice. So in the end, a warning would just force us to do this same book-keeping in two places instead of one.

bbrehm:
OK, supported use-case it is :+1:

PR #415, opened 2024-02-06
help util: replace jtextutils with asciitable

mpollmeier:
* main motivation: jtextutils became vulnerable because the domain
`com.massisframework` is available for sale...
* also: asciitable is better IMO


mpollmeier:
cc @Ferada

traversal/src/main/scala/overflowdb/traversal/help/Table.scala
```diff
@@ -1,21 +1,29 @@
...
+      table.addRule()
+      table.addRow(columnNames.asJava)
+      table.addRule()
+      if (rows.nonEmpty) {
+        rows.map(_.asJava).foreach(table.addRow)
+      }
+      table.addRule()
+      table.getContext.setGridTheme(TA_GridThemes.FULL)
+      table.setTextAlignment(TextAlignment.LEFT)
+      table.render(width)
```
maltek, line 25: what happens when the contents are wider than 120 characters? (e.g. dataflows are wider than that all the time)

mpollmeier: It get's split up into several lines. There's quite a few options, I went for left alignment for the text. 
https://github.com/vdmeer/asciitable/?tab=readme-ov-file#text-alignment-1

mpollmeier: (previously, the table would just expand in width to the widest cell)

mpollmeier:
For comparison: in joern:
old: 
image
image


new:
image
image

mpollmeier:
n.b. I'm working on a followup change that'll pass down the terminal size to the text table

maltek (approved):
ok, that does look nicer :+1:

Ferada:
Thanks for looking into that :+1:

PR #439, opened 2024-07-11
edge property hack: prereq for flatgraph -> odb converter

mpollmeier:
In order to prepare for a potential rollback to overflowdb in the future, we're creating a
flatgraph->overflowdb converter. Since flatgraph supports exactly one edge property only,
there are no edge property names. So when we deserialise and find exactly one edge property with
the name `EDGE_PROPERTY`, and the schema defines exactly one edge property, then we assume that's the one.
Let's be clear: this is a dirty hack, but since we're phasing out overflowdb in favor of flatgraph, it's
arguably ok.


bbrehm (approved):
You can test against https://github.com/joernio/flatgraph/pull/224

mpollmeier:
Looks good!

